PR #411, opened 2015-01-15
CreateSubbuffer support

bastiankoe:
Hi, I added the functionality to create a sub buffer out of a buffer.


kylelutz:
Awesome! Thanks for working on this!

A few small issues though:

* Because `clCreateSubBuffer()` was added in OpenCL 1.1, this needs to be guarded with a `#ifdef CL_VERSION_1_1` check. See the [`event::set_callback()`](https://github.com/kylelutz/compute/blob/master/include/boost/compute/event.hpp#L202) method for an example of how to do this.
* Could you add some documentation for this method? Be sure to include the `\opencl_version_warning` line so users know it is only available for OpenCL 1.1.
* Could you add a small test-case to [`test_buffer.cpp`](https://github.com/kylelutz/compute/blob/master/test/test_buffer.cpp) to test this functionality?
* Also, can you make a pull-request against the `develop` branch? See some documentation for this in the [Developer's Guide](https://github.com/kylelutz/compute/wiki/Developer%27s-Guide) on the wiki.

After that it should be good to merge. Thanks!

bastiankoe:
The new pull request is opened at #412

PR #413, opened 2015-01-16
Register destructor callback is now templated

bastiankoe:
Added a template based register destructor callback function to the memory_object class.

### DIFF
--- a/include/boost/compute/memory_object.hpp
+++ b/include/boost/compute/memory_object.hpp
@@ -16,6 +16,7 @@
 #include <boost/compute/kernel.hpp>
 #include <boost/compute/detail/get_object_info.hpp>
 #include <boost/compute/detail/assert_cl_success.hpp>
+#include <boost/bind.hpp>
 
 namespace boost {
 namespace compute {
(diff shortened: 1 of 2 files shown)


include/boost/compute/memory_object.hpp
```diff
@@ -16,6 +16,7 @@
 #include <boost/compute/kernel.hpp>
 #include <boost/compute/detail/get_object_info.hpp>
 #include <boost/compute/detail/assert_cl_success.hpp>
+#include <boost/bind.hpp>
```
kylelutz: Hmm, I don't think this `#include` is necessary (looks like it's just used by the test)?

kylelutz:
Just one small issue with the extra include. Also, could you rebase this on `develop` as I just merged your other pull-request which also touched `test_buffer.cpp` and now there's a conflict. Thanks!

include/boost/compute/memory_object.hpp
bastiankoe: Yeah it's not necessary. I don't know why I put this there, you're right that this belongs to the test.

bastiankoe:
I hope that now everything is okay ;)

kylelutz:
Merged! Thanks!

Issue #276, opened 2014-10-13
black_scholes test won't work.


anirul:
Was running the test set and found out that black_scholes doesn't work on my laptop.

``` bash
> ./black_scholes
device: HD Graphics 4000
libc++abi.dylib: terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::compute::context_error> >: OpenCL Build Warning : Compiler build log:
<program source>:1:7: warning: no previous prototype for function 'cnd'
float cnd(float d) [...]
```

Seams like the compiler on OSX need predeclaration to work, with predeclaration this seems to work:

``` bash
diff --git a/example/black_scholes.cpp b/example/black_scholes.cpp
index e78e895..952d28d 100644
--- a/example/black_scholes.cpp
+++ b/example/black_scholes.cpp
@@ -69,6 +69,10 @@ int main()

     // source code for black-scholes program
     const char source[] = BOOST_COMPUTE_STRINGIZE_SOURCE(
+       
+               // some compiler need predeclaration
+               float cnd(float d);
+       
         // approximation of the cumulative normal distribution function
         float cnd(float d)
         {
```

Maybe you should not crash on warnings?

Nice work!
##

PR #418, opened 2015-01-25
Modify black scholes example to use internal RNG from library instead of std::rand

roshanr95:
This adresses #276 . Also check if the values are correct... (esp. the put price)
On my setup,
```
$ ./example/black_scholes
device: GeForce GT 650M
option 0 call price: 15.189
option 0 put price: 0
```


kylelutz:
The black-scholes code should still be correct, but different input values will give different output values. Having `0` exactly for the put price is a little weird, I'd try changing the RNG seed value to something else and then use those results.

PR #426, opened 2015-02-22
MSVC 2013 build fixes

jmr1:


kylelutz:
Looks great! Thanks for fixing these!

PR #427, opened 2015-02-23
BUGFIX enable double for key type in radix sort by key

shehzan10:
* pragma khr_fp64 was not being enabled for key type double


kylelutz:
Looks great! Thanks!

If possible, could you add a test-case for this?

shehzan10:
While debugging we found that the crashes were only for buffer sizes greater than 32 and only on NVIDIA devices. AMD and Intel devices were running just fine without the pragma.

I'll try adding a test soon.

PR #440, opened 2015-03-21
added clear_persistent_cache and minor update to persistent caching

salmanulhaq:
Added ```clear_persistent_cache()``` in ```program_cache.hpp``` which removes ```.boost_compute``` directory.

Minor update to persistent caching: instead of computing sha1 hash of just program source, I've also added compiler options and selected device's name so if anything changes, program would be built again.

### DIFF
--- a/include/boost/compute/program.hpp
+++ b/include/boost/compute/program.hpp
@@ -516,7 +516,8 @@ public:
                 << "// " << options << "\n\n"
                 << source;
 
-            hash = detail::sha1(src.str());
+            // Gets rebuilt if either source, or device (indicated by name) or compile options change
+            hash = detail::sha1(src.str() + context.device.name().str() + options.str());
         }
 
         // Try to get cached program binaries:
--- a/include/boost/compute/utility/program_cache.hpp
+++ b/include/boost/compute/utility/program_cache.hpp
@@ -82,6 +82,25 @@ public:
         m_cache.clear();
     }
 
+    /// Clears persistent cache by deleting the boost_compute folder
+    void clear_persistent_cache()
+    {
+#ifdef BOOST_COMPUTE_USE_OFFLINE_CACHE
+        // Path delimiter symbol for the current OS.
+        static const std::string delim = boost::filesystem::path("/").make_preferred().string();
+
+        // Path to appdata folder.
+#ifdef WIN32
+        static const std::string appdata = detail::getenv("APPDATA") 
+            + delim + "boost_compute";
+#else
+        static const std::string appdata = detail::getenv("HOME")
+            + delim + ".boost_compute";
+#endif
+        boost::filesystem::remove_all(appdata);
+#endif
+    }
+
     /// Returns the program object with \p key. Returns a null optional if no
     /// program with \p key exists in the cache.
     boost::optional<program> get(const std::string &key)


include/boost/compute/utility/program_cache.hpp
```diff
@@ -82,6 +82,25 @@ class program_cache : boost::noncopyable
...
     }
 
+    /// Clears persistent cache by deleting the boost_compute folder
+    void clear_persistent_cache()
+    {
+#ifdef BOOST_COMPUTE_USE_OFFLINE_CACHE
+        // Path delimiter symbol for the current OS.
+        static const std::string delim = boost::filesystem::path("/").make_preferred().string();
+
+        // Path to appdata folder.
```
kylelutz: Instead of copying these lines, could you just call `detail::appdata_path()` directly?

kylelutz:
Also, could you open a pull request against the `develop` branch (that's where we do testing/integration before updating the more stable `master` branch). Thanks!

include/boost/compute/utility/program_cache.hpp
```diff
@@ -82,6 +82,25 @@ class program_cache : boost::noncopyable
         m_cache.clear();
     }
 
+    /// Clears persistent cache by deleting the boost_compute folder
+    void clear_persistent_cache()
```
kylelutz: I think this could be a static method (so users can just call `program_cache::clear_persistent_cache()` directly without having to get/create a `program_cache` object).

include/boost/compute/utility/program_cache.hpp
```diff
@@ -82,6 +82,25 @@ class program_cache : boost::noncopyable
...
+
+        // Path to appdata folder.
+#ifdef WIN32
+        static const std::string appdata = detail::getenv("APPDATA") 
+            + delim + "boost_compute";
+#else
+        static const std::string appdata = detail::getenv("HOME")
+            + delim + ".boost_compute";
+#endif
+        boost::filesystem::remove_all(appdata);
```
kylelutz: I can't remember, but will this through an exception if the offline cache directory hasn't been created yet? If so we should catch/ignore that.

include/boost/compute/program.hpp
```diff
@@ -516,7 +516,8 @@ class program
                 << "// " << options << "\n\n"
                 << source;
 
-            hash = detail::sha1(src.str());
+            // Gets rebuilt if either source, or device (indicated by name) or compile options change
+            hash = detail::sha1(src.str() + context.device.name().str() + options.str());
```
kylelutz: As Denis commented, this additional information should already be stored inside the program source and therefore I don't think we need to store it here.

ddemidov:
Regarding @keryell's note: one possible solution is to make `compute::detail::sha1` a functor, where `operator()(const std::string&)` calls `process_bytes()` repeatedly (see the current implementation of `details::sha1` [here](https://github.com/kylelutz/compute/blob/aa15cd6ac4cd76bf7129a0c653dde2e70e8c089c/include/boost/compute/detail/sha1.hpp#L25)). It could also provide a conversion to string operator, so that computing  the hash could be:
```cpp
detail::sha1 hash;
hash(source);
hash(device_name);
hash(compiler_version);

load_program_binary(hash, context);
```

keryell:
Yes you are right, since boost/uuid/sha1.hpp has already the right interface. So there is no need to have a Merkle's tree approach that would create several sha1 intermediate objects.
Great!

PR #441, opened 2015-03-22
Persistent cache

salmanulhaq:
*    Made ```clear_persistent_cache``` static
*    Catching exception if there's no ```.boost_compute``` folder (or in use)
*    Followed @ddemidov suggestion on computing hash for persistent cache lookup

@kylelutz have to re-use ```appdata``` path snippet since ```appdata_path()``` is private within ```program``` but ```clear_persistent_cache``` suits better in ```program_cache```.

### DIFF
--- a/include/boost/compute/utility/program_cache.hpp
+++ b/include/boost/compute/utility/program_cache.hpp
@@ -83,7 +83,7 @@ public:
     }
 
     /// Clears persistent cache by deleting the boost_compute folder
-    void clear_persistent_cache()
+    static void clear_persistent_cache()
     {
 #ifdef BOOST_COMPUTE_USE_OFFLINE_CACHE
         // Path delimiter symbol for the current OS.
@@ -97,7 +97,17 @@ public:
         static const std::string appdata = detail::getenv("HOME")
             + delim + ".boost_compute";
 #endif
-        boost::filesystem::remove_all(appdata);
+        try
+        {
+            if(boost::filesystem::exists(appdata))
+            {
+               boost::filesystem::remove_all(appdata);
+            }  
+        }
+        catch(boost::filesystem::filesystem_error const & err)
+        {
+            std::cout<<err.what()<<std::endl;
+        }
 #endif
     }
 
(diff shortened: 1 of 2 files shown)


ddemidov:
You forgot to actually change the implementation of `boost::compute::detail::sha1`. I went and made the same improvement for my own library [here](https://github.com/ddemidov/vexcl/commit/ab9f0f5a8ed1c294f42b71c64297b66e2c134500); feel free to reuse it.

include/boost/compute/utility/program_cache.hpp
```diff
@@ -82,6 +82,35 @@ class program_cache : boost::noncopyable
         m_cache.clear();
     }
 
+    /// Clears persistent cache by deleting the boost_compute folder
+    static void clear_persistent_cache()
+    {
+#ifdef BOOST_COMPUTE_USE_OFFLINE_CACHE
+        // Path delimiter symbol for the current OS.
+        static const std::string delim = boost::filesystem::path("/").make_preferred().string();
```
kylelutz: So instead of doing all this here, you should call the [`detail::appdath_path()`](https://github.com/kylelutz/compute/blob/develop/include/boost/compute/detail/path.hpp#L30) function defined in `<boost/compute/detail/path.hpp>`. I recently extracted this code from the `program` class so you may have to rebase on top of the `develop` branch to get it.

include/boost/compute/utility/program_cache.hpp
```diff
@@ -82,6 +82,35 @@ class program_cache : boost::noncopyable
...
+#ifdef WIN32
+        static const std::string appdata = detail::getenv("APPDATA") 
+            + delim + "boost_compute";
+#else
+        static const std::string appdata = detail::getenv("HOME")
+            + delim + ".boost_compute";
+#endif
+        try
+        {
+            if(boost::filesystem::exists(appdata))
```
kylelutz: I don't think you need both an existence check and a try/catch block. I'd lean towards just using the if(exists(appdata)) condition and then removing the try/catch.

PR #442, opened 2015-03-22
Basic Threefry support Added

junmuz:
In these commits, basic Threefry RNG support is added.


include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
+// Added By: Muhammad Junaid Muzammil <mjunaidmuzammil@gmail.com>
```
kylelutz: Can you rename this file to `threefry_engine.hpp` in order to match the class name.

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+#include <boost/compute/program.hpp>
+#include <boost/compute/command_queue.hpp>
+#include <boost/compute/algorithm/transform.hpp>
+#include <boost/compute/detail/iterator_range_size.hpp>
+#include <boost/compute/utility/program_cache.hpp>
+#include <boost/compute/iterator/discard_iterator.hpp>
+
+namespace boost {
+namespace compute {
+namespace random {
```
kylelutz: I don't think we need to add an extra namespace for this, just `boost::compute::threefry_engine`.

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+
+/// \class threefry_rng_engine
+/// \brief Threefry pseudorandom number generator.
+template<class T>
```
kylelutz: We could set the default template parameter `T` to `uint_` here (just like in `linear_congruential_engine`).

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+        : m_context(queue.get_context())
+    {
+        // setup program
+        load_program();
+    }
+
+    /// Creates a new threefry_engine object as a copy of \p other.
+    threefry_engine(const threefry_engine<T> &other)
+        : m_context(other.m_context)
+    {
```
kylelutz: I think the copy-constructor also needs to call `load_program()` (or simply copy the program object from `other`).

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /// Copies \p other to \c *this.
+    threefry_engine<T>& operator=(const threefry_engine<T> &other)
+    {
+        if(this != &other){
+            m_context = other.m_context;
```
kylelutz: Same as with the copy-constructor, we need to copy over the program object also.

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+            "    if(Nrounds>30){  X.v[0] += X.v[1]; X.v[1] = RotL_32(X.v[1],R_32x2_6_0); X.v[1] ^= X.v[0]; }\n"
+            "    if(Nrounds>31){  X.v[0] += X.v[1]; X.v[1] = RotL_32(X.v[1],R_32x2_7_0); X.v[1] ^= X.v[0]; }\n"
+            "    if(Nrounds>31){\n"
+            "        X.v[0] += ks[2]; X.v[1] += ks[0];\n"
+            "        X.v[1] += 8;\n"
+            "    }\n"
+            "    return X;\n"
+            "}\n"
+
+            "__kernel void generate_rng(__global uint *ctr, __global uint *key, const uint size) {\n"
```
kylelutz: So this kernel takes in pointers to `uint` arrays, will this work if the `threefry_engine` is setup with a type other than `uint_` (e.g. `ushort_` or `ulong_`)?

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+            "        ctr[i+1] = in.v[1];\n"
+            "    }\n"
+            "}\n";
+
+        m_program = cache->get_or_build(cache_key, std::string(), source, m_context);
+    }
+
+public:
+    template<class OutputIterator>
+    void generate(command_queue &queue, OutputIterator first_ctr, OutputIterator last_ctr, OutputIterator first_key, OutputIterator last_key) {
```
kylelutz: Can you add some documentation for this method? What is the difference between `ctr` and `key`. The other random number engines (`mersenne_twister_engine` and `linear_congruential_engine`) take a single output range of type `T` to write the generated values to. Can you explain why this takes two output ranges?

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+        const size_t size_ctr = detail::iterator_range_size(first_ctr, last_ctr);
+        const size_t size_key = detail::iterator_range_size(first_key, last_key);
+        if(!size_ctr || !size_key || (size_ctr != size_key)) {
+            return;
+        }
+        kernel rng_kernel = m_program.create_kernel("generate_rng");
+        rng_kernel.set_arg(0, first_ctr.get_buffer());
+        rng_kernel.set_arg(1, first_key.get_buffer());
+        rng_kernel.set_arg(2, static_cast<const uint_>(size_ctr));
+        queue.enqueue_1d_range_kernel(rng_kernel, 0, 1, 0);
```
kylelutz: Hmm, so this kernel only executes with a single work-item? Is there any way to parallelize it to better take advantage of massively-parallel GPU devices?

include/boost/compute/random/threefry.hpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+private:
+    context m_context;
+    program m_program;
+};
+
+typedef threefry_engine<uint_> threefry;
```
kylelutz: I don't think we need this `typedef`. I'd just change `threefry_engine` to have a default template argument of `uint_` and then people could just use `threefry_engine<>` (similar to the API for `linear_congruential_engine<>`).

test/CMakeLists.txt
```diff
@@ -173,6 +173,7 @@ add_compute_test("random.bernoulli_distribution" test_bernoulli_distribution.cpp
 add_compute_test("random.discrete_distribution" test_discrete_distribution.cpp)
 add_compute_test("random.linear_congruential_engine" test_linear_congruential_engine.cpp)
 add_compute_test("random.mersenne_twister_engine" test_mersenne_twister_engine.cpp)
+add_compute_test("random.threefry" test_threefry.cpp)
```
kylelutz: Just for consistency, can you name this test `random.threefry_engine` and the source file `test_threefry_engine.cpp`?

test/test_threefry.cpp
```diff
@@ -0,0 +1,98 @@
...
+    rng.discard(5, queue);
+    rng.generate(vector.begin(), vector.end(), queue);
+
+    CHECK_RANGE_EQUAL(
+        uint_, 5, vector,
+        (uint_(4161255391),
+         uint_(3922919429),
+         uint_(949333985),
+         uint_(2715962298),
+         uint_(1323567403))
```
kylelutz: Are these the same numbers that are generated from the `mersenne_twister_engine`? That doesn't seem right to me. Wouldn't the threefry algorithm generate different random numbers?

kylelutz:
Looking good! I've added some review comments. Look them over and let me know what you think. Thanks!

pavanky:
It is possible for three_fry to generate other data types. You can look at how we dot his in ArrayFire: https://github.com/arrayfire/arrayfire/blob/devel/src/backend/opencl/kernel/random.cl

include/boost/compute/random/threefry.hpp
junmuz: Done

include/boost/compute/random/threefry.hpp
junmuz: I have removed the extra namespace. I was thinking that an added namespacing might be beneficial in grouping the random number functionality.

include/boost/compute/random/threefry.hpp
junmuz: Done

include/boost/compute/random/threefry.hpp
junmuz: Yes, I missed it. I have rectified the problem.

include/boost/compute/random/threefry.hpp
junmuz: Done

include/boost/compute/random/threefry.hpp
junmuz: There are other threefry types that I'll be adding in future. Overall there are four types

1. 32x2
2. 32x4
3. 64x2
4. 64x4

I have only added 32x2 up till now, and will be adding the remainder later.

include/boost/compute/random/threefry.hpp
junmuz: The generated random numbers are contained in the "ctr". "key" is only an input range required for generating RNG.

include/boost/compute/random/threefry.hpp
junmuz: I have made necessary corrections to it. Can you take a look at it?

include/boost/compute/random/threefry.hpp
junmuz: I have removed it.

test/CMakeLists.txt
junmuz: I have corrected it.

include/boost/compute/random/threefry_engine.hpp
```diff
@@ -0,0 +1,267 @@
...
+#include <boost/compute/command_queue.hpp>
+#include <boost/compute/algorithm/transform.hpp>
+#include <boost/compute/detail/iterator_range_size.hpp>
+#include <boost/compute/utility/program_cache.hpp>
+#include <boost/compute/iterator/discard_iterator.hpp>
+
+namespace boost {
+namespace compute {
+
+/// \class threefry_rng_engine
```
kylelutz: This should be `\class threefry_engine`.

test/test_threefry_engine.cpp
```diff
@@ -0,0 +1,69 @@
...
+        key[i*2] = 0;
+        key[i*2+1] = 0;
+    }
+
+    boost::compute::copy(ctr, ctr+20, vector_ctr.begin(), queue);
+    boost::compute::copy(key, key+20, vector_key.begin(), queue);
+
+    rng.generate(queue, vector_ctr.begin(), vector_ctr.end(), vector_key.begin(), vector_key.end());
+    CHECK_RANGE_EQUAL(
+        uint_, 20, vector,
```
kylelutz: I think `vector` should actually be `vector_key` here?

include/boost/compute/random/threefry_engine.hpp
```diff
@@ -0,0 +1,267 @@
...
+        m_program = cache->get_or_build(cache_key, std::string(), source, m_context);
+    }
+
+public:
+
+
+    /// Generates Threefry random numbers using both the counter and key values, and then stores
+    /// them to the range [\p first_ctr, \p last_ctr).
+    template<class OutputIterator>
+    void generate(command_queue &queue, OutputIterator first_ctr, OutputIterator last_ctr, OutputIterator first_key, OutputIterator last_key) {
```
kylelutz: So can you explain the difference between the counter and key values? As a user simply wanting to generate random numbers, do I need to set up the counter values myself? The other random number engines use a generate function which just takes a single output range like this:
```
template<class OutputIterator>
void generate(OutputIterator first, OutputIterator last, command_queue &queue)
```
Is it possible to implement the same interface for `threefry_engine`?

include/boost/compute/random/threefry_engine.hpp
junmuz: Comment fixed

test/test_threefry_engine.cpp
junmuz: It was supposed to be vector_ctr. I have fixed it. Tests are passing now.

include/boost/compute/random/threefry_engine.hpp
junmuz: Yes, it can be done. I think we can keep the counter as an optional parameter. We can use a default value which can be overridden by the user. What should be the default value, I think I have to research a bit over it.

kylelutz: Yeah, if possible, it would be great to match the same interface used with the other random-number engines.

junmuz: I have added an overloaded method.

Issue #325, opened 2014-12-03
Add benchmark for find() algorithm


kylelutz:
Add a benchmark for the [`find()`](http://kylelutz.github.io/compute/boost/compute/find.html) algorithm (i.e. `perf_find`).

See the [`perf/perf_count.cpp`](https://github.com/kylelutz/compute/blob/master/perf/perf_count.cpp) benchmark for an example/template.

It would also be good to add a corresponding benchmark for the STL `std::find()` algorithm (i.e. `perf_stl_find`).

haahh:
I may be closed. See https://github.com/kylelutz/compute/pull/444.

kylelutz:
Closed. Thanks!

PR #444, opened 2015-03-26
Benchmarks for find() algorithm #325

haahh:
This addresses issue #325 .

I've added find() algorithm benchmarks for Boost.Compute, STL, Thrust. Benchmarks are based on worst-case scenario (searching for element that isn't in vector).


perf/perf_stl_find.cpp
```diff
@@ -0,0 +1,45 @@
...
+    // create vector of random numbers on the host
+    std::vector<int> host_vector(PERF_N);
+    std::generate(host_vector.begin(), host_vector.end(), rand_int);
+
+    // trying to find element that isn't in vector (worst-case scenario)
+    int wanted = rand_int_max + 1;
+    perf_timer t;
+    for(size_t trial = 0; trial < PERF_TRIALS; trial++){
+        t.start();
+        std::find(host_vector.begin(), host_vector.end(), wanted);
```
kylelutz: So I tried running this version a few times with varying data sizes but always got the same very small time. This usually means the compiler is optimizing out the call the `std::find()` entirely because the result is not actually used. I'd add something to store the result in a variable and compare it against `host_vector.end()`.

kylelutz:
Minor issue with the STL version. After fixing that it should be good to merge. Thanks!

perf/perf_stl_find.cpp
haahh: I don't have that issue. Are performance tests built with -O3 by default? 

Nevertheless, I will commit fix immediately and to be sure it won't happen again with Thrust performance test I will add there the same check.

haahh:
It's interesting that Travis build failed. I'm not sure why.

perf/perf_stl_find.cpp
kylelutz: Not by default, but when I'm running the performance benchmarks and comparing against the STL I enable optimizations (else the results are not very fair). I usually do this by setting `CMAKE_BUILD_TYPE=Release` which enables `gcc -03`.

kylelutz:
Looks great! Merged!

kylelutz:
And don't worry about the `g++-4.8: internal compiler error` Travis build error for GCC. It's not related to your change. I've been looking into it. For now, as long as the clang passes it should be good.

perf/perf_stl_find.cpp
haahh: Thanks! That's exactly what I need to know. From now on I will use that setting when testing performance, especially against STL.

PR #445, opened 2015-03-29
Backwards compatibility

dvdjg:
This pull gives compatibility at runtime with ICDs supporting old OpenCL versions (NVIDIA).
Using of a fast version checking.
Adds support for mapping images.


dvdjg:
Some other changes done.
Changes in some `event` returning functions.

PR #448, opened 2015-04-04
kernel::get_work_group_info() should be const

ddemidov:
Please correct me if I am wrong, but I don't see a reason for `kernel::get_work_group_info()` to be non-const.


kylelutz:
No, you're right, it should be `const`. Thanks for fixing it!

PR #451, opened 2015-04-08
Offline cache optimization

ddemidov:
This gets rid of temporary strings allocation in implementation of offline cache. See #440, #441 for details.


kylelutz:
Looks good, thanks!

Issue #121, opened 2014-05-25
Add skip_iterator class


kylelutz:
Add a `skip_iterator` adaptor which skips over multiple elements each time it is incremented.

For example, the following test should pass:

```
int data[] = { 1, 2, 3, 4, 5, 6, 7, 8 };
compute::vector<int> vec(data, data + 8, queue);

compute::vector<int> result(4, context);

// copy every other element to result
compute::copy(
    compute::skip_iterator<int>(vec.begin(), 2),
    compute::skip_iterator<int>(vec.end(), 2),
    result.begin(),
    queue
);
CHECK_RANGE_EQUAL(int, 4, result, (1, 3, 5, 7));
```

haahh:
I've read about Boost.Iterator (iterator facade and adaptor) and I started coding skip_iterator. 

Sadly, I'm stuck. I can't find an elegant way to modify index expression ```expr.m_index_expr```, so it would include skip parameter. 

```cpp
(...)

template<class InputIterator, class T, class IndexExpr>
struct skip_iterator_index_expr
{
    typedef typename std::iterator_traits<InputIterator>::value_type result_type;

    skip_iterator_index_expr(const InputIterator &input_iter,
                                  const T &skip_value,
                                  const IndexExpr &index_expr)
        : m_input_iter(input_iter),
          m_skip_value(skip_value),
          m_index_expr(index_expr)
    {
    }

    InputIterator m_input_iter;
    const T& m_skip_value;
    IndexExpr m_index_expr;
};

template<class InputIterator, class T, class IndexExpr>
inline meta_kernel& operator<<(meta_kernel &kernel,
                               const skip_iterator_index_expr<InputIterator,
                                                                   T,
                                                                   IndexExpr> &expr)
{
    return kernel << expr.m_input_iter[expr.m_index_expr];
}

(...)
```

Should skip value type be restricted to integers types? Should skip value be only greater than zero?

kylelutz:
I actually happen to have some code which implements this lying around, just never got around to actually fully documenting it and making a pull-request. I decided to name it `strided_iterator` after the one found in the Boost.Range library (see [boost/range/adaptor/strided.hpp](http://www.boost.org/doc/libs/1_58_0/boost/range/adaptor/strided.hpp)).

I think it makes sense to restrict the stride parameter to integral types (I use the `difference_type` from the underlying iterator as the stride parameter type). Also I think it makes sense to limit the stride to values greater than zero.

[...]

PR #452, opened 2015-04-12
Add strided_iterator class (skip_iterator) #121

haahh:
This addresses issue #121.

I've added ```strided_iterator``` adaptor which - as requested - skips over multiple elements each time it is incremented. It's called strided_iterator to match corresponding iterator in the Boost.Range library (see [boost/range/adaptor/strided.hpp](http://www.boost.org/doc/libs/1_58_0/boost/range/adaptor/strided.hpp)). Tests included. 

I've also included function ```make_strided_iterator_end()```. It may be useful. What do you think about it?

@kylelutz, thank you for providing your code. It was helpful.

### DIFF
--- a/include/boost/compute/iterator/strided_iterator.hpp
+++ b/include/boost/compute/iterator/strided_iterator.hpp
@@ -106,7 +106,7 @@ inline meta_kernel& operator<<(meta_kernel &kernel,
 /// \class strided_iterator
 /// \brief Iterator adaptor which skips over multiple elements each time it is incremented.
 ///
-/// TODO
+/// TODO: Description
 ///
 ///
 /// \see buffer_iterator, make_strided_iterator()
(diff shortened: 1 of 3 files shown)


kylelutz:
Looks great! I wonder if we should follow the API in boost/range/adaptor/strided.hpp which has separate `make_begin_strided_iterator()` and `make_end_strided_iterator()` functions? Or alternatively a single `make_strided_iterator()` function which then always takes both the begin and end of the underlying range.

haahh:
In current solution I wanted to be consistent with other iterators in Boost.Compute, which don't really need to know what is below them and where it ends. ```make_strided_iterator_end()``` in this context is just a helper function to find past-the-end element, so caller doesn't have to calculate it.

haahh:
```make_strided_iterator()``` function which takes both the begin and end of the underlying iterator must led to ```strided_iterator``` implementation where increment method would look like this:

```cpp
void increment()
{
   // super_type::base_reference() - underlying iterator referring
   // end - iterator referring to past-the-end element 
   auto step = (std::min)(m_stride, end - super_type::base_reference());
   std::advance(super_type::base_reference(), step);
}
```

Otherwise why we would require end when making strided iterator for begin? And I think only that way we would get strided iterator end for ```make_strided_iterator(end, end, stride)```.

In this version end for ```strided_iterator``` made with ```stride = x``` and ```stride = y``` (where ```x != y```) would be the same. My first thought is that can cause problems when someone want to iterate from the end to the begin.

kylelutz:
That makes sense. Thanks for working on this!

Issue #381, opened 2014-12-29
Add arithmetic operators to valarray


kylelutz:
Add overloads for the arithmetic operators (e.g. `+`, `-`) to `boost::compute::valarray<T>` to match `std::valarray<T>`.

From the review:
```
- <http://kylelutz.github.io/compute/boost/compute/valarray.html> -
presumably compute::valarray has overloaded arithmetic operators?  Are
they mentioned anywhere?
```

guptadivyansh:
http://www.cplusplus.com/reference/valarray/valarray/operators/

Similar to these?

kylelutz:
@guptadivyansh yeah, those are the ones.

guptadivyansh:
https://github.com/guptadivyansh/compute/commit/1771ad07b9dd3f893e2497f6b17ece46ee391c85

@kylelutz Can you please look at this?

kylelutz:
So that will actually perform the operation on the host rather than on the compute device (e.g. GPU). For implementing these functions I'd probably use the [`boost::compute::transform()`](http://kylelutz.github.io/compute/boost/compute/transform.html) algorithm.

archangel4:
Could you provide some more reference to implement this feature ?

kylelutz:
A basic implementation would be something like:
```
valarray<T> operator+(const valarray<T> &other) const
{
    const context &context = m_buffer.get_context();
    command_queue queue(context, context.get_device());
    valarray<T> result(size(), context);
    transform(begin(), end(), other.begin(), plus<T>(), queue);
    queue.finish();
    return result;
}
```

As operator overloads don't allow the command queue argument to be passed, we must create one explicitly to perform the operation. Because of that, actually using these operators would not be ideal for performance. For that reason I'd be hesitant to add them, but they could be useful for their convenience.

haahh:
Logical not operator ``valarray<bool> operator!() const;`` cannot be implemented as in OpenCL it's impossible to build memory buffer with bool type, which makes valarray<bool> incorrect. However, I can implement ``valarray<char> operator!() const;`` operator (or with int) .

kylelutz:
Yeah, I'd just use `valarray<char>`.

kylelutz:
Implemented in PR #454.

PR #454, opened 2015-05-10
Missing valarray operators

haahh:
This addresses issue https://github.com/boostorg/compute/issues/381.

I've implemented all missing valarray operators (see [valarray operators](http://www.cplusplus.com/reference/valarray/valarray/operators/)) and added tests for them.

Implemented operators work only for default context (queue) and are asynchronous.

Note that in OpenCL it's impossible to create memory buffer with bool type and because of that ```valarray<char>``` is used  as a return type for comparison and logical operators instead of ```valarray<bool>```.

### DIFF
--- a/include/boost/compute/container/valarray.hpp
+++ b/include/boost/compute/container/valarray.hpp
@@ -15,6 +15,7 @@
 #include <valarray>
 
 #include <boost/static_assert.hpp>
+#include <boost/type_traits.hpp>
 
 #include <boost/compute/buffer.hpp>
 #include <boost/compute/algorithm/copy.hpp>
@@ -126,7 +127,8 @@ public:
     valarray<T> operator~() const
     {
         BOOST_STATIC_ASSERT_MSG(
-            (is_fundamental<T>::value && !is_floating_point<T>::value),
+            is_fundamental<T>::value &&
+                !is_floating_point<typename scalar_type<T>::type>::value,
             "This operator can be used with all OpenCL built-in scalar"
             " and vector types except the built-in scalar and vector float types"
         );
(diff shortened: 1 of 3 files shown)


haahh:
I fixed some errors I did not encountered in my environment:
* empty macro argument (I did not run into this problem, cause by mistake I was compiling without ```-pedantic -Werror``` options.)
* 'bind' function call error (Somehow bind call works for me correctly without ::boost::compute::)

Now it seems that my valarray tests are to long :/

haahh:
Travis build failed with ```No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.``` error.

On my system all 37 valarray tests work perfectly, both with g++4.8.2 and clang 3.4. I would be grateful if someone would confirm that.

kylelutz:
I can confirm it passes locally on my laptop (Mac OSX, NVIDIA GPU, clang). The failure on Travis is strange, I wish the output had more information.

kylelutz:
But it is good to see how comprehensive your tests are :-).

haahh:
Thanks :-) I try to test all code I commit to Boost Compute.

My best guess now is that Travis was busy and somehow valarray tests were not completed before mentioned ten minutes passed.

haahh:
I checked my code again and I saw I did not add static assert  in unary operators checking if valarray T type is a fundamental type. I will fix it tomorrow morning. (Of course it's not connected to Travis failure.)

haahh:
I will check if the failure is caused by too many tests or some specific test freezes. I think running valarray tests with --log_level=all option would help find the problem, I'll try this when I get home. 

The only difference between my environment and that on Travis (apart from OpenCL drivers) is boost version - I have 1.55, Travis uses 1.48. Sadly I have some odd problems building this version, so I can't recreate exact same environment on my computer.

haahh:
I located the problem - remainder binary operator test. This is [built with remainder binary operator test](https://travis-ci.org/boostorg/compute/jobs/62435729), which fails to complete valarray tests due to ```fatal error in "modulus_binary_operator": signal: integer divide by zero; address of failing instruction: 0x7f82a29a841b```, and this is [built without it](https://travis-ci.org/boostorg/compute/jobs/62435729), that works perfectly (I've temporally modified .travis.yml so it builds only valarray tests - it's faster that way). 

Still, I don't know why this problem occurs. I think test data is correct and IMO this test should work.

kylelutz:
Looks great! Merged!

haahh:
Thanks!

PR #459, opened 2015-05-19
find_extrema based on parallel reduction

haahh:
This adds ```find_extrema_reduce``` function and change ```find_extrema``` function to use it if there is enough local memory on device to perform an efficient parallel reduction. If there is not, ```find_extema_with_atomics``` is used (just like it was).

```find_extrema_reduce``` and ```find_extema_with_atomics``` get the same time results in ```perf_max_element``` benchmark. For vectors with 134217728 elements the difference is visible. 

master branch:
```
=== max_element with stl ===
size,time (ms)
2,0.000231
4,0.000260
8,0.000286
16,0.000299
32,0.000312
64,0.000364
128,0.000421
256,0.000519
512,0.000779
1024,0.001274
2048,0.002192
4096,0.004097
8192,0.007887
16384,0.015551
32768,0.030970
65536,0.061633
131072,0.123388
262144,0.251553
524288,0.506069
1048576,1.010500
2097152,2.298420
4194304,4.540600
8388608,9.029470
16777216,18.179500
33554432,36.464900
=== max_element with compute ===
size,time (ms)
2,2.938670
4,2.953160
8,2.982040
16,3.020810
32,2.960240
64,3.023210
128,3.006550
256,3.052890
512,3.005860
1024,2.995740
2048,3.060540
4096,2.994000
8192,3.076960
16384,3.008020
32768,3.046970
65536,3.040360
131072,3.092490
262144,3.138090
524288,3.187000
1048576,2.986230
2097152,3.074310
4194304,3.280390
8388608,3.608900
16777216,4.278090
33554432,5.508340
--- 
67108864, 7.90407 ms
134217728, 13.3829 ms

=== max_element with bolt ===
size,time (ms)
2,0.239094
4,0.242006
8,0.242288
16,0.243363
32,0.242257
64,0.243461
128,0.243329
256,0.241237
512,0.413800
1024,0.520461
2048,0.729220
4096,1.151180
8192,1.964450
16384,3.618160
32768,6.824630
65536,11.551800
131072,22.584200
262144,28.026900
524288,28.072100
1048576,27.904700
2097152,28.255900
4194304,28.788000
8388608,26.075500
16777216,25.414200
33554432,28.391500
```

pr_improving_find_extrema branch:
```
=== max_element with compute ===
size,time (ms)
2,2.951420
4,3.024390
8,2.964650
16,3.017970
32,3.006210
64,3.144770
128,3.117730
256,3.126830
512,3.064650
1024,3.069900
2048,3.055710
4096,3.098720
8192,3.127860
16384,3.130310
32768,3.098050
65536,3.114520
131072,3.190490
262144,3.240010
524288,3.252840
1048576,3.347880
2097152,3.471760
4194304,3.582500
8388608,3.625560
16777216,4.242310
33554432,5.249320
--- 
67108864, 7.20043 ms
134217728, 11.2549 ms

```

However, when worst case scenario is used, i.e. vector contains elements from 0 to ```vector.size() - 1```, performance of  ```find_extema_with_atomics``` decreases. 

```cpp
int global = 0;
int next_int()
{
    return global++;
}
```

When function ```next_int()``` is used to fill vector instead of ```rand_int()```, performance of ```find_extema_reduce``` remains the same, but performance of  ```find_extema_with_atomics``` decreases:
```
2097152,5.000910
4194304,6.930040
8388608,10.957300
16777216,18.637500
33554432,34.256900
```

Another argument for merging this PR is fact that ```find_extema_reduce``` work for OpenCL 1.0, when current ```find_extema_with_atomics```  does not (I have to say that I don't no why it does not work).

### DIFF
--- a/include/boost/compute/algorithm/detail/find_extrema.hpp
+++ b/include/boost/compute/algorithm/detail/find_extrema.hpp
@@ -12,6 +12,7 @@
 #define BOOST_COMPUTE_ALGORITHM_DETAIL_FIND_EXTREMA_HPP
 
 #include <boost/compute/detail/iterator_range_size.hpp>
+#include <boost/compute/algorithm/detail/find_extrema_reduce.hpp>
 #include <boost/compute/algorithm/detail/find_extrema_with_atomics.hpp>
 #include <boost/compute/algorithm/detail/serial_find_extrema.hpp>
 
@@ -32,9 +33,18 @@ inline InputIterator find_extrema(InputIterator first,
         return first;
     }
 
+    const device &device = queue.get_device();
+
     // use serial method for small inputs
-    if(count < 64){
+    // and when device is a CPU
+    if(count < 64 || (device.type() & device::cpu)){
         return serial_find_extrema(first, last, sign, queue);
+    }
+
+    // find_extrema_reduce() is used only if requirements are met
+    if(find_extrema_reduce_requirements_met(first, last, queue))
+    {
+        return find_extrema_reduce(first, last, sign, queue);
     }
 
     // use serial method for OpenCL version 1.0 due to
--- a/include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
+++ b/include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
@@ -1,0 +1,260 @@
+//---------------------------------------------------------------------------//
+// Copyright (c) 2015 Jakub Szuppe <j.szuppe@gmail.com>
+//
+// Distributed under the Boost Software License, Version 1.0
+// See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt
+//
+// See http://boostorg.github.com/compute for more information.
+//---------------------------------------------------------------------------//
+
+#ifndef BOOST_COMPUTE_ALGORITHM_DETAIL_FIND_EXTREMA_REDUCE_HPP
+#define BOOST_COMPUTE_ALGORITHM_DETAIL_FIND_EXTREMA_REDUCE_HPP
+
+#include <algorithm>
+#include <vector>
+
+#include <boost/compute/types.hpp>
+#include <boost/compute/command_queue.hpp>
+#include <boost/compute/algorithm/copy.hpp>
+#include <boost/compute/container/vector.hpp>
+#include <boost/compute/detail/meta_kernel.hpp>
+#include <boost/compute/detail/iterator_range_size.hpp>
+#include <boost/compute/detail/parameter_cache.hpp>
+#include <boost/compute/memory/local_buffer.hpp>
+#include <boost/compute/type_traits/type_name.hpp>
+#include <boost/compute/utility/program_cache.hpp>
+
+namespace boost {
+namespace compute {
+namespace detail {
+
+template<class InputIterator>
+bool find_extrema_reduce_requirements_met(InputIterator first,
+                                          InputIterator last,
+                                          command_queue &queue)
+{
+    typedef typename std::iterator_traits<InputIterator>::value_type input_type;
+
+    const device &device = queue.get_device();
+
+    // device must have dedicated local memory storage
+    // otherwise reduction would be highly inefficient
+    if(device.get_info<CL_DEVICE_LOCAL_MEM_TYPE>() != CL_LOCAL)
+    {
+        return false;
+    }
+
+    const size_t max_work_group_size = device.get_info<CL_DEVICE_MAX_WORK_GROUP_SIZE>();
+    // local memory size in bytes (per compute unit)
+    const size_t local_mem_size = device.get_info<CL_DEVICE_LOCAL_MEM_SIZE>();
+
+    std::string cache_key = std::string("__boost_find_extrema_reduce_")
+        + type_name<input_type>();
+    // load parameters
+    boost::shared_ptr<parameter_cache> parameters =
+        detail::parameter_cache::get_global_cache(device);
+
+    // Get preferred work group size
+    size_t work_group_size = parameters->get(cache_key, "wgsize", 256);
+
+    work_group_size = (std::min)(max_work_group_size, work_group_size);
+
+    // local memory size needed to perform parallel reduction
+    size_t required_local_mem_size = 0;
+    // indices size
+    required_local_mem_size += sizeof(uint_) * work_group_size;
+    // values size
+    required_local_mem_size += sizeof(input_type) * work_group_size;
+
+    // at least 4 work groups per compute unit otherwise reduction
+    // would be highly inefficient
+    return ((required_local_mem_size * 4) <= local_mem_size);
+}
+
+template<class InputIterator, class ResultIterator>
+inline size_t find_extrema_reduce(InputIterator first,
+                                  size_t count,
+                                  ResultIterator result,
+                                  vector<uint_>::iterator result_idx,
+                                  size_t work_groups_no,
+                                  size_t work_group_size,
+                                  char sign,
+                                  command_queue &queue)
+{
+    typedef typename std::iterator_traits<InputIterator>::value_type input_type;
+
+    const context &context = queue.get_context();
+
+    meta_kernel k("find_extrema_reduce");
+    size_t count_arg = k.add_arg<uint_>("count");
+    size_t output_arg = k.add_arg<input_type *>(memory_object::global_memory, "output");
+    size_t output_idx_arg = k.add_arg<uint_ *>(memory_object::global_memory, "output_idx");
+    size_t block_arg = k.add_arg<input_type *>(memory_object::local_memory, "block");
+    size_t block_idx_arg = k.add_arg<uint_ *>(memory_object::local_memory, "block_idx");
+
+    k <<
+        // Work item global id
+        k.decl<const uint_>("gid") << " = get_global_id(0);\n" <<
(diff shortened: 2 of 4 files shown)


include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
```diff
@@ -0,0 +1,245 @@
...
+    const bool is_queue_ordered =
+        static_cast<bool>(
+            queue.get_properties() &
+            command_queue::enable_out_of_order_execution
+        );
+
+    size_t count = detail::iterator_range_size(first, last);
+    size_t work_group_size = (std::min)(max_work_group_size, preferred_work_group_size);
+    size_t work_groups_no = compute_units_no * 64;
+    work_groups_no = (std::min)(
```
kylelutz: For these sort of calculations, you may be able to use the `calculate_work_size()` function in [`<boost/compute/detail/work_size.hpp>`](https://github.com/boostorg/compute/blob/master/include/boost/compute/detail/work_size.hpp).

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
```diff
@@ -0,0 +1,245 @@
...
+                                  char sign,
+                                  command_queue &queue)
+{
+    typedef typename std::iterator_traits<InputIterator>::value_type input_type;
+
+    const context &context = queue.get_context();
+
+    meta_kernel k("find_extrema_reduce");
+
+    k.add_extension_pragma("cl_amd_printf");
```
kylelutz: Just for debugging?

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
```diff
@@ -0,0 +1,245 @@
...
+    find_extrema_reduce(first, count,
+                        results.begin(), results_idx.begin(),
+                        work_groups_no, work_group_size,
+                        sign,
+                        queue);
+
+    std::vector<input_type> host_results(work_groups_no);
+    std::vector<uint_> host_results_idx(work_groups_no);
+
+    if(!is_queue_ordered) {
```
kylelutz: For unordered queues it is usually preferred to use `event`-based synchronization instead of calling `clFinish()` (which will stop and wait for all kernels to complete, not just those for this algorithm). In this case, you would store the `event` returned from `enqueue_nd_range_kernel()` and pass that as a part of the `wait_list` to the next operation on the queue.

Do you have devices which support out-of-order execution to test with? For the most part, Boost.Compute algorithms assume in-order execution. It may be best to leave this out for now until we can add full out-of-order queue support to Boost.Compute. Let me know what you think.

include/boost/compute/algorithm/detail/find_extrema.hpp
```diff
@@ -32,11 +33,25 @@ inline InputIterator find_extrema(InputIterator first,
...
+
     // use serial method for small inputs
-    if(count < 64){
+    // and when device is a CPU
+    if(count < 64 || (device.type() & device::cpu)){
         return serial_find_extrema(first, last, sign, queue);
     }
 
+    // find_extrema_reduce() is used only if requirements are met
+    size_t preferred_work_group_size = 256;
```
kylelutz: This should probably be read from the `parameter_cache`. For an example, see the [`reduce_on_gpu()`](https://github.com/boostorg/compute/blob/master/include/boost/compute/algorithm/detail/reduce_on_gpu.hpp) algorithm implementation.

kylelutz:
Looks good, left a few inline comments. Benchmark numbers look great!

haahh:
Thanks for all your comments! That's why I like code reviews and four-eyes principle.

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
haahh: Yes, you're right. I know ```event```-based synchronization should be used with out-of-order queue. I just thought "oh, it will not work with out-of-order queue" and I can't get ```event```, so I did what I did :) If we assume in-order, it's unnecessary perform ```finish()``` and indeed it affects other kernels. I will remove this part.

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
haahh: Yeah. I missed it.

include/boost/compute/algorithm/detail/find_extrema.hpp
haahh: OK. In that case I think I will add ```work_group_size```  and ```work_items_per_compute_unit``` parameters, because they both influence performance of the ```find_extrema_reduce```.

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
haahh: I don't know how ```calculate_work_size()``` can help in this situation. It calculate global worki size which I don't need at this point.

I just need to check if work group size isn't greater than maximal work group size for used device, then calculate number of work groups - it is equal to ```compute_units_no * 64``` (64 represents how many values will be reduced by one work item, values per work item/thread), or when vector is small, it is ```static_cast<size_t>(std::ceil(float(count) / work_group_size))```.

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+        k.decl<bool>("compare_result") << ";\n" <<
+        "while( idx < count ){\n" <<
+            // Next element
+            k.decl<input_type>("next") << " = " << first[k.var<uint_>("idx")] << ";\n" <<
+            // Comparison between currently best element (acc) and next element
+            "compare_result = acc " << sign << " next;\n" <<
+            "acc = compare_result ? acc : next;\n" <<
+            "acc_idx = compare_result ? acc_idx : idx;\n" <<
+            "idx += get_global_size(0);\n" <<
```
pavanky: From my experiments, the performance is better if the strides are equal to local size instead of global size. We saw up to 10-20% improvement (atleast for NVIDIA GPUs) in ArrayFire when we made this change. This would require the idx to be computed a bit differently. You can look at the reference code here: https://github.com/arrayfire/arrayfire/blob/devel/src/backend/opencl/kernel/ireduce_first.cl#L11

haahh: Interesting! I'll try to test it. Any idea why there is this 10-20% improvement?

pavanky: I am not entirely sure. My guess would be fewer cache misses. If the data is not aligned to the cache boundary, each read will need to get two cache lines per global stride. If using a local stride, the remaining data from the second cache line will be useful for the second iteration.

haahh: I'm getting the same time results. CodeXL shows the same low cache hit percentage, almost 0.  Maybe I did not understand correctly, but this is how I've changed idx calculation (this is generated kernel code cause it is easier to read):

```cpp
const uint id = (get_group_id(0) * get_local_size(0) * vpwi) + get_local_id(0);
uint m = id + (get_local_size(0) * vpwi);
m = count < m ? count : m;
if(id >= count) {
    return;
}
uint idx = id;
int acc;
uint acc_idx = idx;
acc = _buf0[idx]; // first[k.var<uint_>("idx")]
idx += get_local_size(0);
bool compare_result;
while( idx < m){
    int next = _buf0[idx]; // first[k.var<uint_>("idx")]
    compare_result = acc > next;
    acc = compare_result ? acc : next;
    acc_idx = compare_result ? acc_idx : idx;
    idx += get_local_size(0);
}
...
```
Where ```vpwi``` is values per work item. I have to add that in this kernel number of work groups per compute unit is constant, ```vpwi``` isn't.

pavanky: Let me do a quick test on an nvidia gpu.

pavanky: Is there any chance you can push the alternative version to a different branch so I won't have to reimplement it ?

haahh: Sure, I was just about to do it.

haahh:
I removed out-of-order queue check and printf pragma. I added `parameter_cache` for work group size and number of work groups per compute unit.

kylelutz:
Thanks! Merged!

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
haahh: @pavanky And how does it work? Have you noticed any increase in the performance?

haahh:
Great!

I think after GSoC'15 (or earlier if I will have time) I will add support for custom comparison function for `find_extrema` (therefore for `max_element()` and `min_element()`). I'll create issue for this.

@pavanky Thanks for your input!

include/boost/compute/algorithm/detail/find_extrema_reduce.hpp
pavanky: I have had trouble running the max_element example on my machine. It keeps segfaulting. I don't think it is an issue with your code. It may be a bug in compute itself. I can only test it after the issue is fixed.

PR #462, opened 2015-06-02
Pull Request with Threefry functionality support

junmuz:
Hello,

I think this particular threefry feature can be merged into develop. Right now it only contains a single type, however in future when support will be added for other types the default will still point to the current one (i.e. 32X2). I don't foresee any signature changes for the API in near future. What do you say?


kylelutz:
Looks great! Thanks for all your work on this!

PR #464, opened 2015-06-06
Install BoostComputeConfig.cmake to lib/cmake/BoostCompute

marbre:
I am packaging BoostCompute for Gentoo Linux. Normally on Linux systems cmake configurations go to /usr/lib/cmake, while Find*.cmake modules go to /usr/share/cmale/Modules. I assume the installation path should be changed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ configure_file(
 # install cmake config file
 install(
   FILES ${BoostCompute_BINARY_DIR}/BoostComputeConfig.cmake
-  DESTINATION share/cmake/BoostCompute
+  DESTINATION lib/cmake/BoostCompute
 )
 
 # install header files


kylelutz:
Thanks for fixing this!

ghisvail:
> Normally on Linux systems cmake configurations go to /usr/lib/cmake

Please rephrase this with "on my Linux system", i.e. Gentoo. For other Linux systems like those which are Debian-based this path would be /usr/lib/<multiarch-triplet>/cmake, for arch-based systems it would be /usr/lib32/cmake or /usr/lib64/cmake, and so on...

Please refrain from making such generic assertion @marbre in the future, where all you propose is a patch that addresses *your* distribution specific needs.

Downstream specific changes should stick downstream, period.

Apologize for the dry tone.

ghisvail:
Slight erratum here:

> for arch-based systems it would be /usr/lib32/cmake or /usr/lib64/cmake

For Arch-Linux 32-bits: /usr/lib32
For Arch-Linux 64-bits: /usr/lib
For Fedora 64-bits: /usr/lib64

pavanky:
@ghisvail /usr/lib32 is for 32 bit libraries on 64 bit systems. It is /usr/lib on 32 bit systems.

ghisvail:
My point remains: on Linux, as in "general" Linux, the path where the lib and cmake stuff will eventually land is *downstream specific*. Hence a patch like this serves solves nothing but one specific case.

FYI, I have no particular feeling towards the acceptance of this patch, but justifying it in the name of "generic" Linux is a bit upsetting.

pavanky:
I understand, I just pointed out the 32 bit stuff because you specifically corrected yourself with another comment. Ideally if there was an environment variable that linux distributions used for the library path, we can just use that instead.

ghisvail:
> I just pointed out the 32 bit stuff because you specifically corrected yourself with another comment.

Thanks for correcting me.

marbre:
As far as I get it [FHS](http://www.pathname.com/fhs/pub/fhs-2.3.html#USRLIBLIBRARIESFORPROGRAMMINGANDPA) defines `/usr/lib` as mandatory, while `/usr/lib<qual>` is optional. IMHO on multilib systems `/usr/lib<qual>` linked to `/usr/lib`.
This was not a justification, it was more an explanation. As I do this in my free time , not even needing it for my research, I might sometimes hold it short. Anyway, sorry for the inconveniences.

ghisvail:
> This was not a justification, it was more an explanation.

My apologize for misinterpreting your intentions. 

> As I do this in my free time , not even needing it for my research

That makes two of us.

Issue #374, opened 2014-12-29
Support custom function for inclusive_scan()/exclusive_scan()


kylelutz:
Add support for a custom reduction function to the `inclusive_scan()` and `exclusive_scan()` algorithms.

kylelutz:
Implemented in PR #468.

PR #468, opened 2015-06-19
Support for custom functions in inclusive and exclusive scan

haahh:
This addresses issue https://github.com/boostorg/compute/issues/374.

Now in inclusive and exclusive scan custom associative binary operator can be used instead of default addition operator. For exclusive scan also initial value can be set. 

I've updated docs for inclusive and exclusive scan.

I've also added single argument constructor for vector types (``int2(1) == int2(1,1) == vector_type<int,2>(1)``, ``int4(4) = int4(4, 4, 4, 4)  == vector_type<int,4>(1)``, etc.).

### DIFF
--- a/include/boost/compute/algorithm/detail/scan_on_cpu.hpp
+++ b/include/boost/compute/algorithm/detail/scan_on_cpu.hpp
@@ -23,11 +23,13 @@ namespace boost {
 namespace compute {
 namespace detail {
 
-template<class InputIterator, class OutputIterator>
+template<class InputIterator, class OutputIterator, class T, class BinaryOperator>
 inline OutputIterator scan_on_cpu(InputIterator first,
                                   InputIterator last,
                                   OutputIterator result,
                                   bool exclusive,
+                                  T init,
+                                  BinaryOperator op,
                                   command_queue &queue)
 {
     if(first == last){
@@ -36,15 +38,32 @@ inline OutputIterator scan_on_cpu(InputIterator first,
 
     typedef typename
         std::iterator_traits<InputIterator>::value_type input_type;
+    typedef typename
+        std::iterator_traits<OutputIterator>::value_type output_type;
 
     const context &context = queue.get_context();
 
     // create scan kernel
     meta_kernel k("scan_on_cpu");
-    k.add_arg<ulong_>("n");
+
+    // Arguments
+    size_t n_arg = k.add_arg<ulong_>("n");
+    size_t init_arg = k.add_arg<output_type>("initial_value");
+
+    if(!exclusive){
+        k <<
+            k.decl<const ulong_>("start_idx") << " = 1;\n" <<
+            k.decl<output_type>("sum") << " = " << first[0] << ";\n" <<
+            result[0] << " = sum;\n";
+    }
+    else {
+        k <<
+            k.decl<const ulong_>("start_idx") << " = 0;\n" <<
+            k.decl<output_type>("sum") << " = initial_value;\n";
+    }
+
     k <<
-        k.decl<input_type>("sum") << " = 0;\n" <<
-        "for(ulong i = 0; i < n; i++){\n" <<
+        "for(ulong i = start_idx; i < n; i++){\n" <<
         k.decl<const input_type>("x") << " = "
             << first[k.var<ulong_>("i")] << ";\n";
 
@@ -52,7 +71,9 @@ inline OutputIterator scan_on_cpu(InputIterator first,
         k << result[k.var<ulong_>("i")] << " = sum;\n";
     }
 
-    k << "    sum = sum + x;\n";
+    k << "    sum = "
+        << op(k.var<output_type>("sum"), k.var<output_type>("x"))
+        << ";\n";
 
     if(!exclusive){
         k << result[k.var<ulong_>("i")] << " = sum;\n";
@@ -65,7 +86,8 @@ inline OutputIterator scan_on_cpu(InputIterator first,
 
     // setup kernel arguments
     size_t n = detail::iterator_range_size(first, last);
-    scan_kernel.set_arg<ulong_>(0, n);
+    scan_kernel.set_arg<ulong_>(n_arg, n);
+    scan_kernel.set_arg<output_type>(init_arg, static_cast<output_type>(init));
 
     // execute the kernel
     queue.enqueue_1d_range_kernel(scan_kernel, 0, 1, 1);
--- a/include/boost/compute/types/fundamental.hpp
+++ b/include/boost/compute/types/fundamental.hpp
@@ -53,6 +53,13 @@ public:
 
     vector_type()
     {
+
+    }
+
+    explicit vector_type(const Scalar scalar)
+    {
+        for(int i = 0; i < N; i++)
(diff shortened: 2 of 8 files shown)


include/boost/compute/algorithm/detail/scan_on_cpu.hpp
```diff
@@ -36,15 +38,32 @@ inline OutputIterator scan_on_cpu(InputIterator first,
...
     meta_kernel k("scan_on_cpu");
-    k.add_arg<ulong_>("n");
+
+    // Arguments
+    size_t n_arg = k.add_arg<ulong_>("n");
+    size_t init_arg = k.add_arg<output_type>("init");
+
+    if(!exclusive){
+        k <<
+            k.decl<const ulong>("start_idx") << " = 1;\n" <<
```
kylelutz: Use Boost.Compute's `ulong_` instead of `ulong` (which is not always available).

include/boost/compute/algorithm/detail/scan_on_cpu.hpp
```diff
@@ -36,15 +38,32 @@ inline OutputIterator scan_on_cpu(InputIterator first,
...
+            k.decl<output_type>("sum") << " = " << first[0] << ";\n" <<
```
kylelutz: It's usually better to not directly embed user data inside the kernel source, this leads to kernel recompilation for each different input range. Rather it's better to either pass it as a kernel argument or look it up in the input range inside the kernel.

include/boost/compute/algorithm/detail/scan_on_cpu.hpp
```diff
@@ -36,15 +38,32 @@ inline OutputIterator scan_on_cpu(InputIterator first,
...
+            result[0] << " = sum;\n";
+    }
+    else {
+        k <<
+            k.decl<const ulong>("start_idx") << " = 0;\n" <<
```
kylelutz: Same here, `ulong_`.

include/boost/compute/algorithm/detail/scan_on_cpu.hpp
```diff
@@ -63,9 +84,12 @@ inline OutputIterator scan_on_cpu(InputIterator first,
     // compile scan kernel
     kernel scan_kernel = k.compile(context);
 
+//    std::cout << k.source() << std::endl;
```
kylelutz: Just for debugging?

include/boost/compute/types/fundamental.hpp
```diff
@@ -53,6 +53,13 @@ class vector_type
 
     vector_type()
     {
+
+    }
+
+    vector_type(const Scalar scalar)
```
kylelutz: We should probably mark this `explicit` to prevent unintentional conversions from scalars to vectors.

include/boost/compute/types/fundamental.hpp
```diff
@@ -102,12 +109,18 @@ class vector_type
...
     m_value[i] = BOOST_PP_CAT(arg, i);
+#define BOOST_COMPUTE_VECTOR_TYPE_ASSIGN_CTOR_SINGLE_ARG(z, i, _) \
+    m_value[i] = arg;
 
 #define BOOST_COMPUTE_DECLARE_VECTOR_TYPE_CLASS(cl_scalar, size, class_name) \
     class class_name : public vector_type<cl_scalar, size> \
     { \
     public: \
         class_name() { } \
+        class_name( scalar_type arg ) \
```
kylelutz: Same here, probably mark this `explicit`.

test/test_scan.cpp
```diff
@@ -175,4 +180,160 @@ boost::compute::exclusive_scan(
...
+int data[] = { 1, 2, 1, 2, 3 };
+boost::compute::vector<int> input(data, data + 5, queue);
+
+// setup output
+boost::compute::vector<int> output(5, context);
+
+// inclusive scan with multiplication
+boost::compute::inclusive_scan(
+    input.begin(), input.end(), output.begin(),
+    boost::compute::multiplies<int>()
```
kylelutz: You should pass the same `queue` argument here.

test/test_scan.cpp
```diff
@@ -175,4 +180,160 @@ boost::compute::exclusive_scan(
...
+
+    BOOST_CHECK_EQUAL(input.size(), size_t(5));
+    BOOST_CHECK_EQUAL(output.size(), size_t(5));
+
+    CHECK_RANGE_EQUAL(int, 5, output, (1, 2, 2, 4, 12));
+
+    // in-place inclusive scan
+    CHECK_RANGE_EQUAL(int, 5, input, (1, 2, 1, 2, 3));
+    boost::compute::inclusive_scan(input.begin(), input.end(), input.begin(),
+                                   boost::compute::multiplies<int>());
```
kylelutz: Same here, make sure you pass the `queue` argument so that the calls to `inclusive_scan` and the code to check the values are all on the same `command_queue`.

test/test_scan.cpp
```diff
@@ -175,4 +180,160 @@ boost::compute::exclusive_scan(
...
+boost::compute::vector<int> input(data, data + 5, queue);
+
+// setup output
+boost::compute::vector<int> output(5, context);
+
+// exclusive_scan with multiplication
+// initial value equals 10
+boost::compute::exclusive_scan(
+    input.begin(), input.end(), output.begin(),
+    int(10), boost::compute::multiplies<int>()
```
kylelutz: Pass `queue` here.

kylelutz:
Looks great! Left a few inline comments.

haahh:
Thanks for pointing out all these mistakes! I really don't know how - once again - I didn't delete that kernel source print, I thought I did :) 

Fixed, tested and pushed.

test/test_scan.cpp
```diff
@@ -175,4 +180,164 @@ boost::compute::exclusive_scan(
...
+{
+    size_t size = 1000;
+    bc::vector<int> device_vector(size, int(2), queue);
+    BOOST_CHECK_EQUAL(device_vector.size(), size);
+    bc::inclusive_scan(device_vector.begin(), device_vector.end(),
+                       device_vector.begin(), bc::multiplies<int>(), queue);
+
+    std::vector<int> host_vector(size, 2);
+    BOOST_CHECK_EQUAL(host_vector.size(), size);
+    bc::copy(device_vector.begin(), device_vector.end(), host_vector.begin());
```
kylelutz: You should use `queue` here too (and for the rest of the algorithm calls in this test). In general, the tests should always use the context and command queue set up in `context_setup.hpp` instead of falling back to the default queue.

kylelutz:
Just a few places to add the queue parameter in the test and this should be good to go. Thanks!

test/test_scan.cpp
haahh: OK, thanks!

I've found out that the context or the queue parameter is missing in a few tests of other algorithms (e.g., test_accumulate.cpp, test product_int) . Should I fix it?

haahh:
Fixed. I also added the queue parameter in previous tests as it was missing there too.

kylelutz:
Merged! Thanks!

PR #470, opened 2015-06-24
Added CUDADIR to scaning paths

tpasternak:
Hello

As CUDA is not always installed at /usr/lib64/nvidia it is reasonable to scan CUDADIR environment variable

Regards,
Tomek


kylelutz:
Looks good. Thanks!

PR #475, opened 2015-07-02
Add missing queue and context parameters in tests

haahh:
This adds missing queue and context parameters to functions calls in tests. I've also changed some tests to use queue, context and device from `context_setup.hpp`.


kylelutz:
Awesome! Thanks for fixing these!

PR #478, opened 2015-07-16
Fixing problems reported in Boost Inspection Report

haahh:
I fixed some of problems reported in [Boost Inspection Report for 1.59 beta rc1](http://boost.cowic.de/rc/docs-inspect-develop.html#compute). I could not fix all of them due to problems/issues/doubts I described below.

I fixed violations of Boost min/max guidelines, converted tabs into spaces, added missing license information and copyrights (based on who and when created the file). I'm not sure how to correctly add missing license and copyright info to .qbk files. 

I fixed one non-ASCII character in `/example/opencv_convolution.cpp`, but I did not change names with non-ASCII character/s as I don't think it would be right.

A few of Boost marcos used in Boost.Compute are deprecated, some since 1.50, some since 1.51. Boost.Compute requires Boost in version 1.48 or later. New macros that replaced deprecated ones (see [Boost macros](http://www.boost.org/doc/libs/1_58_0/libs/config/doc/html/boost_config/boost_macro_reference.html)) are - of course - not present in 1.48, so if Boost.Compute should work with 1.48 we need to keep these deprecated macros. Otherwise, required Boost version should be set to 1.51. 

( I also check that in `compute/include/boost/compute/config.hpp` internal macros like BOOST_COMPUTE_DETAIL_NO_VARIADIC_TEMPLATES are defined, but still sometimes in the code original macros from Boost are used, so that should be fixed too? )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,12 @@
+# ---------------------------------------------------------------------------
+#  Copyright (c) 2013 Kyle Lutz <kyle.r.lutz@gmail.com>
+# 
+#  Distributed under the Boost Software License, Version 1.0
+#  See accompanying file LICENSE_1_0.txt or copy at
+#  http://www.boost.org/LICENSE_1_0.txt
+#
+# ---------------------------------------------------------------------------
+
 cmake_minimum_required(VERSION 2.8)
 
 project(BoostCompute)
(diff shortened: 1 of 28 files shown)


kylelutz:
Thanks for fixing all these! Could you make this pull request against the `develop` branch?

haahh:
Oh, sorry for this mistake :-) I'll add more fixes today and make pull
request against develop branch.

Could you comment doubts/problems I listed?
18 lip 2015 08:09 "Kyle Lutz" <notifications@github.com> napisał(a):

PR #480, opened 2015-07-18
Fixing problems reported in Boost Inspection Report

haahh:
I fixed some of problems reported in [Boost Inspection Report for 1.59 beta rc1](http://boost.cowic.de/rc/docs-inspect-develop.html#compute). 

* I fixed violations of Boost min/max guidelines, converted tabs into spaces, added missing license information and copyrights (based on who and when created the file).
* I fixed one non-ASCII character in `/example/opencv_convolution.cpp`, but I _DID NOT_ change author name with non-ASCII character/s as I don't think it would be right.
* I added `BOOST_NO_CXX11_VARIADIC_TEMPLATES` macro which replaces deprecated `BOOST_NO_VARIADIC_TEMPLATES`.
* I also moved all `BOOST_NO_*` to `config.hpp`. 
* I used `BOOST_COMPUTE_NO_` prefix for all these macros, but it's not a problem to changed the prefix if different is preferred (for example  `BOOST_COMPUTE_DETAIL_NO_`).
* (!) The last commit removes deprecated Boost macros and update required version of Boost Library to 1.55 (in Boost 1.48 new macros that replaces deprecated ones are not defined, theoretically 1.51 is enough, but IMO 1.55 is currently the most popular one + there are no packages for 1.51). This commit can be removed if we want to keep supporting 1.48 version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,12 @@
+# ---------------------------------------------------------------------------
+#  Copyright (c) 2013 Kyle Lutz <kyle.r.lutz@gmail.com>
+# 
+#  Distributed under the Boost Software License, Version 1.0
+#  See accompanying file LICENSE_1_0.txt or copy at
+#  http://www.boost.org/LICENSE_1_0.txt
+#
+# ---------------------------------------------------------------------------
+
 cmake_minimum_required(VERSION 2.8)
 
 project(BoostCompute)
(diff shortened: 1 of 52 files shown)


kylelutz:
Looks good! As for your last comment, I think we still want to keep support for Boost 1.48 (at least until Boost.Compute is in the official Boost release).

haahh:
OK, so I will remove last commit and it will be ready to merge.

haahh:
Done, last commit is removed. Deprecated macros will still generate problems in Boost inspection program, but we keep support for 1.48 version. My commit will wait patiently ;)

PR #487, opened 2015-08-01
Fix overlooked violation of the Boost min/max guidelines

haahh:
This fixes violation of the Boost min/max guidelines I missed in https://github.com/boostorg/compute/pull/480. 

There is new Boost Inspection Report (28.07): http://boost.cowic.de/rc/docs-inspect-develop.html#compute.


kylelutz:
Merged! Thanks!

Issue #476, opened 2015-07-10
Reduction failure on MacOSX with Intel CPU as OpenCL device


fpvidal:
With Intel CPU devices on Mac OS X: Doesn't properly support OpenCL __local memory when the work-group size is greater than one.

Below are the results of test_reduce:

 /usr/local/share/compute/bin/test->./test_reduce
Running 12 test cases...
/usr/local/share/compute/include/boost/compute/command_queue.hpp:1040: fatal error in "boost::compute::event boost::compute::command_queue::enqueue_nd_range_kernel(const boost::compute::kernel &, size_t, const size_t *, const size_t *, const size_t *, const boost::compute::wait_list &)": std::exception: Invalid Work Group Size
/usr/local/share/compute/include/boost/compute/command_queue.hpp:1040: fatal error in "boost::compute::event boost::compute::command_queue::enqueue_nd_range_kernel(const boost::compute::kernel &, size_t, const size_t *, const size_t *, const size_t *, const boost::compute::wait_list &)": std::exception: Invalid Work Group Size
/usr/local/share/compute/test/test_reduce.cpp:48: last checkpoint
-- Repeat-fill case
/usr/local/share/compute/include/boost/compute/command_queue.hpp:1040: fatal error in "boost::compute::event boost::compute::command_queue::enqueue_nd_range_kernel(const boost::compute::kernel &, size_t, const size_t *, const size_t *, const size_t *, const boost::compute::wait_list &)": std::exception: Invalid Work Group Size
/usr/local/share/compute/test/test_reduce.cpp:48: last checkpoint
/usr/local/share/compute/include/boost/compute/command_queue.hpp:1040: fatal error in "boost::compute::event boost::compute::command_queue::enqueue_nd_range_kernel(const boost::compute::kernel &, size_t, const size_t *, const size_t *, const size_t *, const boost::compute::wait_list &)": std::exception: Invalid Work Group Size
/usr/local/share/compute/test/test_reduce.cpp:48: last checkpoint
/usr/local/share/compute/include/boost/compute/command_queue.hpp:1040: fatal error in "boost::compute::event boost::compute::command_queue::enqueue_nd_range_kernel(const boost::compute::kernel &, size_t, const size_t *, const size_t *, const size_t *, const boost::compute::wait_list &)": std::exception: Invalid Work Group Size
/usr/local/share/compute/test/test_reduce.cpp:112: last checkpoint
[...]

PR #491, opened 2015-08-08
Using serial_reduce() when reducing on CPU device

haahh:
There is a bug which causes using `reduce_on_gpu` whenever reduction function is `plus<T>`, even if device is not a GPU. Now for CPU devices `serial_reduce` is always used. 

This should fix https://github.com/boostorg/compute/issues/476.

Also this change improves performance of reduce algorithm as `reduce_on_gpu` is not efficient on CPU. 

```
Device: AMD Phenom(tm) II X3 710 Processor (CPU)

=== accumulate with compute === [ MASTER, reduce_on_gpu is used]
size,time (ms)
2,0.179798
4,0.186708
8,0.172284
16,0.187684
32,0.163798
64,0.185535
128,0.173303
256,0.186923
512,0.168256
1024,0.173956
2048,0.200156
4096,0.235499
8192,0.282913
16384,0.664291
32768,0.699428
65536,1.259930
131072,2.511260
262144,4.673290
524288,9.389420
1048576,17.946400
2097152,38.079900
4194304,76.685300
8388608,123.566000
16777216,296.414000
33554432,603.888000

=== accumulate with compute === [ FIX, serial_reduce is used ]
size,time (ms)
2,0.075449
4,0.075187
8,0.076612
16,0.074292
32,0.073260
64,0.072319
128,0.072425
256,0.077508
512,0.068213
1024,0.077216
2048,0.074187
4096,0.077526
8192,0.083586
16384,0.087925
32768,0.103001
65536,0.149885
131072,0.201135
262144,0.307403
524288,0.557656
1048576,1.031420
2097152,2.347310
4194304,4.742430
8388608,9.330300
16777216,18.533900
33554432,36.776200
```
(This is `perf_accumulate` but accumulate uses reduce algorithm in this benchmark.)

### DIFF
--- a/include/boost/compute/algorithm/reduce.hpp
+++ b/include/boost/compute/algorithm/reduce.hpp
@@ -198,16 +198,6 @@ inline void generic_reduce(InputIterator first,
     }
 }
 
-template<class InputIterator, class T>
-inline void dispatch_reduce(InputIterator first,
-                            InputIterator last,
-                            const buffer_iterator<T> result,
-                            const plus<T> &function,
-                            command_queue &queue)
-{
-    reduce_on_gpu(first, last, result, function, queue);
-}
-
 template<class InputIterator, class OutputIterator, class T>
 inline void dispatch_reduce(InputIterator first,
                             InputIterator last,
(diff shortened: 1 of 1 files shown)


haahh:
https://github.com/boostorg/compute/issues/476 can be closed (I assume you tested it on OSX like you said you'd do).

Issue #365, opened 2014-12-29
Return events from all command_queue::enqueue_* methods


kylelutz:
Currently all of the asynchronous methods in the `command_queue` class return `event` objects. For consistency, we should also return events from the synchronous methods (like `enqueue_write_buffer()`). These events can be useful for chaining together other operations or performance profiling.
##

PR #496, opened 2015-08-16
Return events from all command_queue::enqueue_* methods

haahh:
This resolves https://github.com/boostorg/compute/issues/365. However, there're two methods that needed workarounds:

1. `clEnqueueBarrier(...)`  in OpenCL 1.0 and 1.1 does not take event argument, so obviously event object cannot be returned in universal `enqueue_barrier(void)` method, but there is already `event enqueue_barrier(const wait_list &events)` method so user can just call it with empty list of events to get event associated with enqueued barrier. 
2. I think that in mapping methods like `enqueue_map_buffer` it's better to return pointer to mapped region like it's now. Therefore, instead of modifying them to return an event object I added event argument to these methods.

I also added missing enqueue_map_image method and asynchronous version of a few other `command_queue` methods.

### DIFF
--- a/include/boost/compute/command_queue.hpp
+++ b/include/boost/compute/command_queue.hpp
@@ -249,16 +249,18 @@ public:
     /// \see_opencl_ref{clEnqueueReadBuffer}
     ///
     /// \see copy()
-    void enqueue_read_buffer(const buffer &buffer,
-                             size_t offset,
-                             size_t size,
-                             void *host_ptr,
-                             const wait_list &events = wait_list())
+    event enqueue_read_buffer(const buffer &buffer,
+                              size_t offset,
+                              size_t size,
+                              void *host_ptr,
+                              const wait_list &events = wait_list())
     {
         BOOST_ASSERT(m_queue != 0);
         BOOST_ASSERT(size <= buffer.size());
         BOOST_ASSERT(buffer.get_context() == this->get_context());
         BOOST_ASSERT(host_ptr != 0);
+
+        event event_;
 
         cl_int ret = clEnqueueReadBuffer(
             m_queue,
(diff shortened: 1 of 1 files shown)


kylelutz:
Looks good. I need to think about the new function signatures a little bit.

haahh:
Ok. I can move last two commits to another pull-request if you're ready to accept  https://github.com/haahh/compute/commit/9ec5dcaca56f3b13013356b3a74b8a859313da07 (first commit) which resolves https://github.com/boostorg/compute/issues/365.

kylelutz:
Yeah, let's do that.

My main concern is the new `const map_buffer_event&` arguments which are actually written to (as an output argument) event though they are `const`. Perhaps this would be better as either a pointer defaulted to `0` or have separate overloads and accept a non-const reference. I'm not sure though. I'll think more about it.

haahh:
Yeah, you're right. This change is already introduced in the first commit https://github.com/haahh/compute/commit/9ec5dcaca56f3b13013356b3a74b8a859313da07 - `const event &event` argument was added to `enqueue_map_buffer` function (in next commit I renamed it to `const event &map_buffer_event ` and moved before `const wait_list &events` argument). Second point describes why I did this that way. Now this change is gone.

I lean toward the solution with a non-const reference (both for synchronous and asynchronous methods). 

```cpp
    void* enqueue_map_buffer(const buffer &buffer,
                             cl_map_flags flags,
                             size_t offset,
                             size_t size,
                             event& map_buffer_event,
                             const wait_list &events = wait_list())
    {
        ...
    }
  
    ///\overload
    void* enqueue_map_buffer(const buffer &buffer,
                             cl_map_flags flags,
                             size_t offset,
                             size_t size,
                             const wait_list &events = wait_list())
    {
        event event_;
        return enqueue_map_buffer(buffer, flags, offset, size, event_, events); 
    }
```

PR #503, opened 2015-09-08
Fix install destination of headers

marbre:
Without this fix the headers get, at least on my system, installed to `/usr/include/compute/boost/compute` instead of `/usr/include/boost/compute`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,4 +109,4 @@ install(
 )
 
 # install header files
-install(DIRECTORY include/boost DESTINATION include/compute)
+install(DIRECTORY include/boost DESTINATION include)


marbre:
`BoostComputeConfig.cmake.in` needs to be modified to if this patch is accepted. Before doing so, I would appreciate any feedback.

kylelutz:
Yeah, this is the expected behavior. For now, you should install to `/usr/include/compute` and then pass `-I/usr/include/compute` to include the Boost.Compute headers. Once in the standard Boost release (should be soon now), Boost.Compute will be installed with the rest of the Boost libraries under `/usr/include/boost`.

marbre:
Thanks for clarifying. Closing this pull request, as this is the expected behaviour.

Issue #505, opened 2015-09-08
system::default_device() fails to respect BOOST_COMPUTE_DEFAULT_DEVICE_TYPE for non gpu devices


pisto:
This is because the type of the device id is a bitmask, but the test used are equalities. This means that if the device type is cpu but it is also default, it's type will be `CL_DEVICE_TYPE_DEFAULT|CL_DEVICE_TYPE_CPU != boost::compute::device::cpu`.

haahh:
What do you mean by that? You cannot set CPU as default device? Because for me `export BOOST_COMPUTE_DEFAULT_DEVICE_TYPE=CPU` works. 

Can you point the file and line where `device.type()` is compared to `boost::compute::device::cpu / gpu / accelerator` instead of `&`?

haahh:
Did you mean this [include/boost/compute/system.hpp#L229](https://github.com/boostorg/compute/blob/master/include/boost/compute/system.hpp#L229)?

pisto:
It may work for you if your CPU device is not the default one. I don't know how to mark system-wide my devices as defaults. Anway, I can find these occurrencies:
```
include/boost/compute/algorithm/count_if.hpp:41:    if(device.type() & device::cpu){
include/boost/compute/algorithm/reduce.hpp:175:    if(device.type() & device::cpu){
include/boost/compute/algorithm/detail/scan.hpp:33:    if(device.type() & device::cpu){
include/boost/compute/algorithm/detail/find_extrema.hpp:40:    if(count < 64 || (device.type() & device::cpu)){
include/boost/compute/system.hpp:229:                    if (!(device.type() & device::gpu))
include/boost/compute/system.hpp:233:                    if (!(device.type() & device::cpu))
include/boost/compute/system.hpp:250:            if(device.type() & device::gpu){
include/boost/compute/system.hpp:259:            if(device.type() & device::cpu){
test/test_command_queue.cpp:166:       device.type() == boost::compute::device::gpu){
example/list_devices.cpp:31:            if(device.type() & compute::device::gpu)
example/list_devices.cpp:33:            else if(device.type() & compute::device::cpu)
example/list_devices.cpp:35:            else if(device.type() & compute::device::accelerator)
```

I suggest you clear CL_DEVICE_TYPE_DEFAULT inside device::type().

pisto:
Updated comment, a few times, see previous.

haahh:
Aaaaa, OK, now I get it.

haahh:
[...]

PR #506, opened 2015-09-08
fix a few instances where the device type is not used as a bit mask

pisto:
See https://github.com/boostorg/compute/issues/505


pisto:
I Travis is failing because the compiler goes OOM.

kylelutz:
Looks good! Thanks for fixing these!

Could you make the pull-request against the `develop` branch? After that it should be good to merge. Thanks!

PR #508, opened 2015-09-11
fix a few instances where the device type is not used as a bit mask

pisto:
See https://github.com/boostorg/compute/issues/505


haahh:
Thanks!

Issue #245, opened 2014-08-20
Intermittent test_search failure on Travis-CI


kylelutz:
The following test failure occurs occasionally on Travis-CI:
```
test_search: /home/travis/build/kylelutz/compute/test/test_search.cpp:45: void compute_test::search_int::test_method(): Assertion `iter == vectort.begin() + 10' failed.
Running 2 test cases...
unknown location(0): fatal error in "search_int": signal: SIGABRT (application abort requested)
*** 1 failure detected in test suite "TestSearch"
```

See the build here: https://travis-ci.org/kylelutz/compute/jobs/33041316

We should investigate this (potentially run it on the Intel CPU OpenCL implementation under valgrind).

haahh:
I got almost the same problem (but in `test_search_n.cpp`) with clang on Travis-CI (https://travis-ci.org/boostorg/compute/builds/79994943). I check it on my PC (AMD HD7770) on `master` and this error occurs both with clang and g++.  It does not happen every time.

```
Running 2 test cases...
test_search_n: /home/jakub/git/compute/test/test_search_n.cpp:37: void compute_test::search_int::test_method(): Assertion `iter == vectort.begin() + 10' failed.
unknown location(0): fatal error in "search_int": signal: SIGABRT (application abort requested)

*** 1 failure detected in test suite "TestSearchN"

```

haahh:
OK, I know where the bug is.  I'll push a fix in a few minutes.

haahh:
This was fixed in https://github.com/boostorg/compute/pull/511. Can be closed.

kylelutz:
Closed. Thanks!

PR #510, opened 2015-09-12
Events for queue enqueue_map_ methods

haahh:
This addresses https://github.com/boostorg/compute/issues/365 (https://github.com/boostorg/compute/pull/496 only partiality fixed that issue). Now `enqueue_map_buffer()` returns associated event through non-const parameter `map_buffer_event` (there is also overload without this parameter).
 
I also added missing enqueue_map_image method (with test in `test_image2d.cpp`) and asynchronous versions for enqueue_map_buffer(), enqueue_read_buffer_rect() and enqueue_write_buffer_rect().

### DIFF
--- a/include/boost/compute/command_queue.hpp
+++ b/include/boost/compute/command_queue.hpp
@@ -606,12 +606,15 @@ public:
     #endif // CL_VERSION_1_2
 
     /// Enqueues a command to map \p buffer into the host address space.
+    /// Event associated with map operation is returned through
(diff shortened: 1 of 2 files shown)


haahh:
This https://github.com/boostorg/compute/issues/245 bug occurred in Travic-CI run with clang, but it's not connected to this PR (this bug also occurs on master, soon it'll be fixed).

kylelutz:
Tested this out on my machines, seems to work well. Thanks!

PR #511, opened 2015-09-12
Fix for search and search_n algorithm

haahh:
This addresses https://github.com/boostorg/compute/issues/245.

I found the bug:

If you analyse [search_n](https://github.com/boostorg/compute/blob/master/include/boost/compute/algorithm/search_n.hpp) (check lines 52, 53, 79), you'll see that it puts 1 or 0 in the first `iterator_range_size(t_first, t_last) + 1 - n` positions (that happens in kernel), but later first occurrence of value 1 is searched in range `[matching_indices.begin(); matching_indices.end())` (line 121). The same bug is in `search`.

I fixed it by reducing the size of `matching_indices` to `iterator_range_size(t_first, t_last) + 1 - n`  and adding if clause for cases when pattern does not occur in input vector.

I also added tests to make sure that the pattern gets found even when it occurs at the end of the input vector.

### DIFF
--- a/include/boost/compute/algorithm/search.hpp
+++ b/include/boost/compute/algorithm/search.hpp
@@ -26,7 +26,7 @@ namespace compute {
 ///
 /// Searches for the first match of the pattern [p_first, p_last)
 /// in text [t_first, t_last).
-/// \return Iterator pointing to beginning of first occurence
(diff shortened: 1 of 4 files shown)


kylelutz:
Awesome! Merged!

PR #520, opened 2015-09-28
scatter if operation

jpola:
Hi,

In my project I needed the scatter if operation so I thought I will also push it here. Could you please check it if this is useful?

Thanks,


include/boost/compute/algorithm/scatter_if.hpp
```diff
@@ -0,0 +1,119 @@
+//---------------------------------------------------------------------------//
+// Copyright (c) 2013 Kyle Lutz <kyle.r.lutz@gmail.com>
```
haahh: You should put your name here ;)

test/test_scatter_if.cpp
```diff
@@ -0,0 +1,141 @@
+//---------------------------------------------------------------------------//
+// Copyright (c) 2013 Kyle Lutz <kyle.r.lutz@gmail.com>
```
haahh: and here!

kylelutz:
Looks great, thanks for adding this!

One issue with initialization in the tests (from the clang build in travis-ci):
```
[ 42%] Building CXX object test/CMakeFiles/test_search.dir/test_search.cpp.o
/home/travis/build/boostorg/compute/test/test_scatter_if.cpp:38:21: error: 
      non-aggregate type 'bc::vector<int>' cannot be initialized with an
      initializer list
    bc::vector<int> output = {-1, -1, -1, -1, -1, -1, -1, -1, -1, -1};
                    ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

However, you should be able to change this to use the size and constant constructor like so:
```
bc::vector<int> output(input.size, -1, queue);
```

kylelutz:
Merged! Thanks!

PR #521, opened 2015-10-02
Skip fill() tests with double if not supported by device

kylelutz:
This skips the tests for fill() and fill_n() with double values
if the device does not support the cl_khr_fp64 extension.


kylelutz:
@haahh This fixes some issues with `test_fill` on my laptop (the GPU doesn't support `double` in kernels). Are there other places we should do the same check?

haahh:
I think `double` is used only in `test_fill` and your changes fixes it. I'm not sure, but maybe in some tests we should check if the devices supports atomic operations.

Issue #523, opened 2015-10-04
Any allocator, except buffer_allocator, inhibit move constructor of vector


pisto:
A `boost::compute::vector` that does not use `boost::compute::buffer_allocator` (as for example `boost::compute::pinned_allocator`) can't use its move constructor. Inspect with a debugger the actual constructor that is called here:
```
compute::vector<int, compute::pinned_allocator<int>> a(100);
compute::vector<int, compute::pinned_allocator<int>> b(move(a));   //copy constructor
compute::vector<int> c(100);
compute::vector<int> d(move(c));   //move constructor effectively called
```
I get this behavior with g++ 5.1.1. Actually ,gdb shows for the `b(move(a))` call the first line of the class when stepping into the constructor, which is usually what happens when the default synthesized copy constructor is called.

pisto:
I found the problem. There are several places in vector.hpp where you declare other vector arguments as `vector<T>`. This is different from type `vector` within the class scope, because the `Alloc` template still takes the default value, which is in fact `buffer_allocator`. See this [ideone](https://ideone.com/fUfvNa). This means that any function with this mistake incurs in a copy through a default synthesized copy constructor, if an allocator different than `buffer_allocator` is used. I should say also that this is not just a performance problem (e.g. you get a copy construction but then everything runs fine), because I kept getting crashes, probably because of some misplaced `clRetain*` and `clRelease*` calls.

I fear that I have seen this kind of mistake somewhere in the codebase, and a proper fixing will require some thought on how to handle operations between vectors with different allocators, so I leave patch crafting to you. For now, I use this for vector.hpp: https://gist.github.com/pisto/dd4f0dea133e03e618ba

haahh:
I confirm. PR in process.

haahh:
I think it can be closed, see https://github.com/boostorg/compute/pull/525.

pisto:
Should be fixed. Please also consider auditing copy construction or assignment, it crashed for some reason. Also there's no clear definition of what happens when moving/assigning between vectors with different allocators. The STL tends to have an overly complicate approach on these things imo, so I don't think it's a very good idea to follow it.

haahh:
@pisto In current implementation (after merging https://github.com/boostorg/compute/pull/525) you can not use move operator between vectors with different allocators. Nut sure if it's needed. Assignment operator works correctly between vectors with different allocators.

[...]

PR #525, opened 2015-10-05
Vector with custom allocator fix

haahh:
See https://github.com/boostorg/compute/issues/523. 

In the first commit I added test for vector's move constructor when custom allocator is used (in that case `boost::computer::pinned_allocator<T>`).  You can checkout to this commit and see that test fails on `BOOST_CHECK(a.size() == 0)` and `BOOST_CHECK(a.get_buffer().get() == 0)` as the copy constructor is used instead of the move ctor.  It also crashes because of some memory access violation that I can not locate right now.

Second commit adds more tests for vector when used with custom allocator.

The last commit includes fixes @pisto presented in https://github.com/boostorg/compute/issues/523 + I added `queue` parameter to copy constructor and a copy constructor for situations when different allocators are used.

### DIFF
--- a/include/boost/compute/container/vector.hpp
+++ b/include/boost/compute/container/vector.hpp
@@ -188,14 +188,29 @@ public:
     }
 
     /// Creates a new vector and copies the values from \p other.
-    vector(const vector<T> &other)
+    vector(const vector &other,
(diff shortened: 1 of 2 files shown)


kylelutz:
Awesome! Thanks for fixing this!

Issue #528, opened 2015-10-24
command_queue::enqueue_svm_memcpy(..) does not work


haahh:
Hi,

Tests `fill_svm_buffer`, `copy_svm_ptr` and `sum_svm_kernel` fail on my AMD Radeon R9 380. I managed to figure out that is all because of `command_queue::enqueue_svm_memcpy(..)` not working. It does not copy the data. When map and unmap operations are used instead of memcpy, everything works.

It looks like AMD drivers has a bug in `clEnqueueSVMMemcpy()`, but I'll have to test it without Boost.Compute. When it's confirmed, I'll do workaround for this (or we'll just wait for fix). 

For now I'll push pull request fixing `fill_svm_buffer` and `sum_svm_kernel`, because in those cases copy operations are just a part of preparations for testing (not tested algorithm) and also because not working `command_queue::enqueue_svm_memcpy(..)` is not the only bug in `sum_svm_kernel` test.

Environment: 
```
OS: Linux Mint 17.2 (3.16)
Platform Version: OpenCL 2.0 AMD-APP (1800.11)
Board name: AMD Radeon (TM) R9 380 Series 
Device OpenCL C version: OpenCL C 2.0 
AMD GPU Drivers: 15.9 (15.201.1151)
```

kylelutz:
Hmm, that is frustrating. IIRC, `clEnqueueSVMMemcpy()` worked fine for me with both AMD and Intel OpenCL 2.0 drivers. What happens on your machine, just no data is ever copied? One interesting thing to check is whether copying different amounts of data takes different amounts of time. This could let us know if it's simply unimplemented (if it always just returns and takes essentially zero time) or if it is doing something, just not correctly.

In any case, it would probably be good to add this to `test/quirks.hpp` and still keep a test or two for `enqueue_svm_memcpy()` around for devices which support it correctly.

haahh:
I opened a discussion on AMD forum - https://community.amd.com/message/2680309#2680309, but so far nobody has answered me. I attached short program that tests `clEnqueueSVMMemcpy()`.

OK, I tested the execution time. It turned out that if event is assigned to  `clEnqueueSVMMemcpy()` operation `clGetEventProfilingInfo()` returns error `CL_PROFILING_INFO_NOT_AVAILABLE -7`. After this I tried measuring time with `std::chrono` I always get 0 ms.

[...]

PR #529, opened 2015-10-24
Fix OpenCL 2.0 test cases in test_fill and test_svm_ptr

haahh:
See #528.

I replaced broken `enqueue_svm_memcpy()` operations with map, copying and unmap in `test_fill.cpp` and `test_svm_ptr.cpp`.

I also added required `-cl-std=CL2.0` build option in `sum_svm_kernel` test case ([clBuildProgram](https://www.khronos.org/registry/cl/sdk/2.0/docs/man/xhtml/clBuildProgram.html): *Applications are required to specify the –cl-std=CL2.0 option if they want to compile or build their programs with OpenCL C 2.0.*").

This branch is based on develop, because otherwise there are conflicts.

### DIFF
--- a/test/quirks.hpp
+++ b/test/quirks.hpp
@@ -28,8 +28,17 @@ inline bool is_pocl_device(const boost::compute::device &device)
 // AMD platforms have a bug when using struct assignment. this affects
 // algorithms like fill() when used with pairs/tuples.
 //
-// see: http://devgurus.amd.com/thread/166622
+// see: https://community.amd.com/thread/166622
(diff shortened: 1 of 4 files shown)


haahh:
I've pushed two more commits

Now test case copy_svm_ptr is skipped on AMD devices (see #528 if you want to know why). I also added more tests for enqueue_svm_memcpy() for devices which support it correctly.

kylelutz:
Looks good, thanks for adding it to `quirks.hpp`.

Issue #532, opened 2015-11-05
Build failure on Windows


pisto:
These lines should have `uint_` instead of `uint`: https://github.com/boostorg/compute/blob/master/include/boost/compute/algorithm/detail/find_if_with_atomics.hpp#L81-L83

haahh:
Can be closed - see https://github.com/boostorg/compute/pull/533.

Issue #530, opened 2015-11-04
CL_MAP_WRITE_INVALIDATE_REGION is a valid flag in 1.2


pisto:
No need for this ifdef https://github.com/boostorg/compute/blob/master/include/boost/compute/command_queue.hpp#L88

haahh:
Based on this https://www.khronos.org/registry/cl/sdk/1.1/docs/man/xhtml/enums.html#cl_map_flags and this https://www.khronos.org/registry/cl/sdk/1.2/docs/man/xhtml/enums.html#cl_map_flags, I think ifdef is needed, but it should be `CL_VERSION_1_2` instead of `CL_VERSION_2_0`.

pisto:
Yeah.

haahh:
Thanks ;) I'll fix it soon.

haahh:
Can be closed - see https://github.com/boostorg/compute/pull/533.

PR #533, opened 2015-11-05
Minor fixes

haahh:
This fixes https://github.com/boostorg/compute/issues/532 and https://github.com/boostorg/compute/issues/530.


kylelutz:
Awesome! Thanks for the quick fix!

PR #534, opened 2015-11-18
program::create_with_source_file doesn't fail with bad input

olivas:
Calling boost::compute::program::create_with_source_file silently creates a program if the input filename is incorrect.  Even program.build() succeeds.  This patch throws std::ios_base::failure if stream.fail() is true.  This seems preferable to something else failing way downstream because of a typo in the filename.

I added a test, but didn't actually test it since I'm using this outside the standard boost build system.  I think it should work, but it's easy to figure out what it does.  It simply tries to create a program from a source file, where the filename is an empty string.  The test verifies that this throws.

### DIFF
--- a/include/boost/compute/program.hpp
+++ b/include/boost/compute/program.hpp
@@ -394,6 +394,10 @@ public:
         // open file stream
         std::ifstream stream(file.c_str());
 
+        if(stream.fail()){
+          BOOST_THROW_EXCEPTION(std::ios_base::failure("failed to create stream."));
+        }
+
         // read source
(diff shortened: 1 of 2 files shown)


haahh:
Hi, I will check and test it tomorrow. Could you make this pull request against the `develop` branch?

haahh:
Also, check clang build of your pull request on Travis (don't worry about g++ internal compiler error, it's a recurrent problem).

kylelutz:
Closed in favor of PR #535.

PR #542, opened 2015-12-18
Fix incorrect count update in rngs

haahh:
In `generate` methods in `mersenne_twister_engine`, `linear_congruential_engine` and `threefry_engine` classes there are incorrect updates of count variables. They leads to assigning out-of-range in used kernels. (`#include <algorithm> ` is added for `std::min`.)

In `linear_congruential_engine` there is incorrect `seed` method call: there is no `seed` method with just one parameter of type result_type.


kylelutz:
Thanks for catching these. Merged!

PR #544, opened 2015-12-26
Modify mersenne_twister_engine to copy its states.

okdshin:
The copy ctor and assign op of compute::mersenne_twister_engine does not copy its states. So copied engine causes "Invalid Program" error.
This modification solves this problem.


kylelutz:
Thanks for fixing this!

It would probably be good to use the same pattern for the other random number engines.

PR #545, opened 2015-12-27
Modify linear_congruential_engine to copy its all states.

okdshin:
The copy ctor and the assign op of compute::linear_congruential_engine does not copy its all states. So copied engine causes "Invalid Program" error.
This modification solves this problem.


haahh:
Thanks for the fix! Works for me :+1: 

IMO when fixing it's good to make 2 commits - add tests in the 1st commit and fix in the 2nd. That way it's easy to check (using just git) that it didn't work correctly and now it works fine.

okdshin:
Thanks for your advice, haahh. I fixed so :+1:

kylelutz:
Merged! Thanks!

PR #547, opened 2016-01-05
Fix missing includes.

Flast:
see [compute - test_svm_ptr](http://www.boost.org/development/tests/develop/developer/output/Flast-FreeBSD10-gcc-5-2-0~gnu++11-boost-bin-v2-libs-compute-test-test_svm_ptr-test-gcc-5-2-0-debug.html) and [compute - test_copy](http://www.boost.org/development/tests/develop/developer/output/Flast-FreeBSD10-gcc-5-2-0~gnu++11-boost-bin-v2-libs-compute-test-test_copy-test-gcc-5-2-0-debug.html)

### DIFF
--- a/test/test_copy.cpp
+++ b/test/test_copy.cpp
@@ -16,6 +16,7 @@
 #include <string>
 #include <sstream>
 #include <iterator>
+#include <iostream>
 
 #include <boost/compute/svm.hpp>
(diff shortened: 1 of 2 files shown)


kylelutz:
Thanks for the fix!

PR #560, opened 2016-03-01
Configure documentation to use boost paths.

danieljames:
Thie fixes the stylesheet and navigation images at http://www.boost.org/doc/libs/master/libs/compute/doc/html/index.html


kylelutz:
Awesome! Thanks for the fix!

PR #562, opened 2016-03-04
Fix FAQ section regarding differences with existing libraries

ddemidov:
Mention that VexCL is in fact may use Boost.Compute as a backend.


kylelutz:
Thanks!

Issue #563, opened 2016-03-06
-Wno-deprecated-declarations can not be in Jamfile.v2


haahh:
CXX option `-Wno-deprecated-declarations` can not be a requirement in Jamfile.v2 because this option is not recognizable by MSVC and some other compilers (`/wd4996` should be used for MSVC in this case).

I would fix it myself but I'm not familiar with Boost.Build tool.

http://www.boost.org/development/tests/develop/developer/output/oracle-intel-S2-12-4-stlport4-boost-bin-v2-libs-compute-test-test_accumulate-test-sun-12-4_stlport4-release-threading-multi.html

http://www.boost.org/development/tests/develop/developer/output/teeks99-08c-win2012R2-64on64-boost-bin-v2-libs-compute-test-test_accumulate-test-msvc-10-0-dbg-adrs-mdl-64-thrd-mlt.html

kylelutz:
Good catch, we should fix this. If I recall there is a way to include or exclude specific compilers which we could probably use for this, I'll see what I can find.

haahh:
Everything is described here: https://svn.boost.org/trac/boost/wiki/Guidelines/WarningsGuidelines

It's also possible to do this in code with:

```cpp
#if defined(BOOST_MSVC)
  #pragma warning(push)
  #pragma warning(disable: 4996)
#endif
#include <cl.h> // and other standard OpenCL headers
#if defined(BOOST_MSVC)
  #pragma warning(pop)
#endif
```
.. and accordingly for gcc and clang.

PR #566, opened 2016-03-07
Update tests compile options in Jamfile.v2

haahh:
This is a fast fix for https://github.com/boostorg/compute/issues/563 (+ I've suppressed some other warning for MSVC that are just annoying) . 

I'll later remove lines

```
<toolset>gcc:<cxxflags>-Wno-deprecated-declarations # Deprecated declarations in CL/cl.h
<toolset>clang:<cxxflags>-Wno-deprecated-declarations # Deprecated declarations in CL/cl.h
<toolset>darwin:<cxxflags>-Wno-deprecated-declarations # Deprecated declarations in CL/cl.h
<toolset>msvc:<cxxflags>/wd4996 # Deprecated declarations in CL/cl.h
```
and suppress deprecated declarations warning (related to CL/cl.h) in the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
     - POCL_COMPILER=clang++-3.7
     # Global build options and C++ flags
     - CMAKE_OPTIONS="-DBOOST_COMPUTE_BUILD_TESTS=ON -DBOOST_COMPUTE_BUILD_EXAMPLES=ON -DBOOST_COMPUTE_BUILD_BENCHMARKS=ON -DBOOST_COMPUTE_USE_OFFLINE_CACHE=ON -DBOOST_COMPUTE_ENABLE_COVERAGE=ON -DBOOST_COMPUTE_HAVE_OPENCV=ON -DBOOST_COMPUTE_THREAD_SAFE=ON"
-    - CXX_FLAGS="-Wall -pedantic -Werror -Wno-variadic-macros -Wno-long-long -Wno-shadow -Wno-deprecated-declarations"
+    - CXX_FLAGS="-Wall -pedantic -Werror -Wno-variadic-macros -Wno-long-long -Wno-shadow"
     # Misc
     - RUN_TESTS=true
 
--- a/example/opencl_test.cpp
+++ b/example/opencl_test.cpp
@@ -8,6 +8,62 @@
 // See http://boostorg.github.com/compute for more information.
 //---------------------------------------------------------------------------//
 
+// See boost/compute/detail/diagnostic.hpp
+// GCC
+#if ((__GNUC__ * 100) + __GNUC_MINOR__) >= 402
+#define BOOST_COMPUTE_GCC_DIAG_STR(s) #s
+#define BOOST_COMPUTE_GCC_DIAG_JOINSTR(x,y) BOOST_COMPUTE_GCC_DIAG_STR(x ## y)
+# define BOOST_COMPUTE_GCC_DIAG_DO_PRAGMA(x) _Pragma (#x)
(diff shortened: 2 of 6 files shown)


haahh:
* I've suppressed those deprecated declarations warnings in the code. 
* I have also added OpenCL link flag for FreeBSD. If you take a look on [Boost regression](http://www.boost.org/development/tests/develop/developer/compute.html) you'll see that for some FreeBSD there is a link error, not a compilation error => there is "CL/cl.h", so maybe there is OpenCL implementation too.

kylelutz:
Cool! I like the idea of doing this in the code rather than build system files.

However, I think we can do this in a slightly cleaner way. Currently this approach requires defining some global macros (like `GCC_DIAG_ON`) in our headers and then surrounding our potentially-deprecated OpenCL calls with them (which is quite a few lines of boilerplate which may also have to be modified/updated in the future as compilers change).

A nicer approach may be defining a pair of headers, `disable_warnings.hpp` which disable all the diagnostics and and `enable_warnings.hpp` which re-enables the diagnostics. The disable header would be included at the top of each of our headers that call (potentially deprecated) OpenCL functions, and the enable header would be included at the bottom to reenable the warnings (as to not affect any other header/source files which include our headers). There are a few examples of this in Boost, for instance Boost.Random and Boost.Endian both take similar approaches.

haahh:
I like this idea with a pair of headers, but disabling warning for the whole file? IIRC [command_queue.hpp](https://github.com/boostorg/compute/blob/master/include/boost/compute/command_queue.hpp) has ~1700 lines, it's big enough that in the future deprecated OpenCL function can be used by a mistake. With warning disabled for the whole file, there would be no sign of that.

I can reduce it to two macros (one before, one after deprecated OpenCL function) or to one macro (deprecated OpenCL function as an argument).

haahh:
Moreover, IMHO it's hard to notice that the warnings are disabled for the whole file and in this option with headers there is no information in the code why those warnings are disabled. What do you think?

kylelutz:
Yeah I agree, disabling them for the whole file isn't ideal. If we could get this down to a two-liner push-pop, or a one-liner function wrapper macro (similar to `BOOST_COMPUTE_ASSERT_CL_SUCCESS`) I think that would be better.

haahh:
Unfortunately, one-liner function wrapper macro somehow does not work in `g++` (maybe related to [53431](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431); it works in clang 3.7), so I've done a two-liner.

one-liner (does not work):
```cpp
// This does not work in g++ (4.8)

// Suppress deprecated-declarations warnings for code
#define BOOST_COMPUTE_DEPRECATED_DECLARATIONS_OFF(code) \
    BOOST_COMPUTE_DIAG_OFF(deprecated-declarations, deprecated-declarations, 4996) \
    code \
    BOOST_COMPUTE_DIAG_ON(deprecated-declarations, deprecated-declarations, 4996)
```

two-liners usage:
```cpp
BOOST_COMPUTE_DIAG_OFF(sign-compare, sign-compare, 4018);
a < b
BOOST_COMPUTE_DIAG_ON(sign-compare, sign-compare, 4018);

// for deprecated declarations warning it's simplified:
BOOST_COMPUTE_DISABLE_DEPRECATED_DECLARATIONS();
.. // deprecated functions
BOOST_COMPUTE_ENABLE_DEPRECATED_DECLARATIONS();
```

I want to add that macros related to suppressing deprecated declarations warning are also included in `opencl_test.cpp` because it should be free of any dependency on Boost.Compute.

kylelutz:
Looks good! Thanks!

PR #568, opened 2016-03-08
Fixed nth_element, added unary_nor

koosha94:
This commit adds unary_nor, and nor1 as a logical operator.
It also adds/fixes support for user defined types in nth_element.

### DIFF
--- a/include/boost/compute/algorithm/nth_element.hpp
+++ b/include/boost/compute/algorithm/nth_element.hpp
@@ -43,8 +43,7 @@ inline void nth_element(Iterator first,
             first, last, ::boost::compute::bind(compare, _1, value), queue
         );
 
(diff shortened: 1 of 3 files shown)


haahh:
Thanks for that! It's a good idea. I have some comments:
* NOR is a binary operator (negation of OR). 
* NOR operator can be constructed using `binary_negate` and `logial_or`:

```cpp
not2(logical_or(arg1, arg2))
```

* Don't use `master` branch for pull requests, make a special feature branch (in this case it could be called `fix-nth-element' or whatever), because now your `master` and Boost.Compute are different and that can make problems when you want to make next pull request. 
* It would be great to have test for user-defined types for nth_element.

Maybe it would good to add `find` algorithm version that works with user defined types (`Compare` or `Equal` binary function would be required)?

PR #573, opened 2016-03-16
Add stable_sort_by_key()

haahh:
This PR adds `stable_sort_by_key()`. Currently, `stable_sort_by_key()` works on GPUs only for
keys that are `radix-sortable` and for very small inputs (serial insertion sort algorithm).


kylelutz:
Awesome! Merged!

pavanky:
@kylelutz When is the next release expected ? We want to include a properly defined way to identify if the functionality is available in a given install of boost compute. If we can tie this down to a version it'd be great.

> Currently, stable_sort_by_key() works on GPUs only for
keys that are radix-sortable and for very small inputs (serial insertion sort algorithm).

@haahh Do you mean it has to be both radix sortable and is limited to small sizes or is it either one or the other ?

haahh:
Either one or the other. Basically, Boost.Compute radix sort is stable so we used it here, but we do not have comparison sort implementation yet.

PR #575, opened 2016-03-18
Provide array subscript operator for wait_list

ddemidov:

PR #576, opened 2016-03-18
Wait list improvements

ddemidov:
This merges #574, #575, and also adds wait-list iterators for completeness.


ddemidov:
wrong branch

PR #580, opened 2016-04-07
Add missing return to svm_ptr assignment operator

kylelutz:


haahh:
I wonder if we need to find a stable POCL version (commit), use it instead of master branch because it became too unstable and just upgrade it from time to time after testing that new version is stable.

kylelutz:
Yeah, I think that would be better and make the CI results more dependable. Would it make sense to stick to POCL releases or other tagged versions?

haahh:
I'm not sure about releases (0.12 is 600 commits behind master,  0.13 - 200 commits behind) I will have to test them. If 0.13 works I'll use it, otherwise I'll just figure out the latest commit that works. I'll do on this on Saturday.

kylelutz:
Sounds good to me, thanks!

PR #585, opened 2016-04-17
Add initializer list constructor to wait_list

bwitherspoon:
Constructing a wait_list from an initializer list is often convenient.


kylelutz:
Looks great! Thanks!

PR #592, opened 2016-04-23
Fix bug in optical flow example

haahh:
Image sampler should be a global constant in OpenCL program. See [this](https://groups.google.com/forum/#!topic/boost-compute/RWvaCsEIcL4).


kylelutz:
Looks good, thanks!

Issue #588, opened 2016-04-21
stable_sort_by_key is not stable when greater<T>() is used


pavanky:
- `sort_by_key` uses `radix_sort_by_key` which only supports `less<T>()`.
- `sort_by_key` for `greater<T>()` [was implemented](https://github.com/boostorg/compute/pull/345) by simply reversing both the outputs from `less<T>()`. Which makes it **not** stable.
- `stable_sort_by_key` just wraps `sort_by_key` thereby not being stable when `greater<T>` is called.

pavanky:
One possible hack would be to do the following:

1) Negate the input keys
2) Sort in ascending order
3) Negate the output keys.

haahh:
Good points. Hmm.. Negating the input keys in the algorithms of radix_sort is a good idea (quick and cheap in terms of performance), but it won't work for unsigned keys. We need to locate the place in the radix_sort algorithm where we can switch something and get descending order.

pavanky:
> Negating the input keys in the algorithms of radix_sort is a good idea (quick and cheap in terms of performance), but it won't work for unsigned keys. We need to locate the place in the radix_sort algorithm where we can switch something and get descending order.

Wouldn't the same place where you are doing float --> int transform for sort on floats be ideal ? I think you just need to have another ifdef for [radix](https://github.com/boostorg/compute/blob/master/include/boost/compute/algorithm/detail/radix_sort.hpp#L95) function for descending order.

> On the other hand, we can just change unsigned values so 0 becomes MAX_UINT etc. Would that work?

This is [exactly what we are doing](https://github.com/shehzan10/arrayfire/blob/sort/src/backend/opencl/kernel/sort_by_key_impl.hpp#L85) on our end as a work around for now.

haahh:
Ok. Thanks. Currently, I'm abroad but I'll look into it on Sunday or Monday.
22 kwi 2016 09:36 "Pavan Yalamanchili" <notifications@github.com>
napisał(a):

haahh:
This can be closed. Fixed by https://github.com/boostorg/compute/pull/593.

PR #593, opened 2016-04-23
Fix radix_sort_by_key for desc order

haahh:
This should fix https://github.com/boostorg/compute/issues/588.

Now in radix sort when `greater<T>()` is specified as comparison function (desc order) instead of performing asc-ordered radix sort and then reversing the result, descending radix sort is performed.

It's done by negating input value if its type is a signed type and by subtracting the input value from the maximum value of value's type in other cases (in `radix()` function). That way values seem "reversed" and we get desc order. It should also be faster than sorting in asc order and then reversing the result.

I also added some test for desc-ordered radix sort and radix sort by key.


pavanky:
looks good :+1:

Issue #587, opened 2016-04-21
boost::compute::normal_distribution variates have wrong standard deviation, and can produce infinite/nan values


jagerman:
The normal_distribution implementation in boost::compute has some problems.

First, it generates values with standard deviation about 1.2 times what was requested.  The following code demonstrates:

```
#include <boost/compute/system.hpp>
#include <boost/compute/command_queue.hpp>
#include <boost/compute/algorithm/accumulate.hpp>
#include <boost/compute/algorithm/inner_product.hpp>
#include <boost/compute/container/vector.hpp>
#include <boost/compute/random/default_random_engine.hpp>
#include <boost/compute/random/normal_distribution.hpp>
#include <boost/compute/lambda.hpp>

int main() {
    boost::compute::device device(boost::compute::system::default_device());
    boost::compute::context context(boost::compute::system::default_context());
    boost::compute::command_queue queue(boost::compute::system::default_queue());
    boost::compute::vector<float> vec(1000000, context);
    boost::compute::default_random_engine engine(queue);
    boost::compute::normal_distribution<float> distribution(5.0f, 1.0f);
    distribution.generate(vec.begin(), vec.end(), engine, queue);
    float mean = boost::compute::accumulate(vec.begin(), vec.end(), 0.f) / vec.size();
    float mean_sq = boost::compute::inner_product(vec.begin(), vec.end(), vec.begin(), 0.f) / vec.size();
    float stdev = sqrt(mean_sq - mean*mean);
    std::cout << "mean: " << mean << " (expect 5); stdev: " << stdev << " (expect 1)\n";
}
```

which outputs:

mean: 5.00037 (expect 5); stdev: 1.20036 (expect 1)

This is caused by these lines in boost/compute/random/normal_distribution.hpp:
```
            const RealType z1 = sqrt(-2.f * log2(x1)) * cos(2.f * M_PI_F * x2);
            const RealType z2 = sqrt(-2.f * log2(x1)) * sin(2.f * M_PI_F * x2);
```
both of those log2 calls have to be log (i.e. using the natural logarithm).  The ~1.2 scale in the standard deviation comes out of this mathematically: log2(x) = ln(x) / ln(2), and the square root of 1/ln(2) equals 1.2011224.  i.e. the standardized variates are too large by a factor of sqrt(1/ln(2)).

I *think* (but have not verified) that there is another problem here: the generated x1 and x2 values can take on values of either 0 or a value larger than 1:
```
[...]

PR #595, opened 2016-04-24
Fix uniform real distribution 

haahh:
Related to https://github.com/boostorg/compute/issues/587. 

Now generated floating-point values are uniformly distributed on the interval `[a, b)` and no value is greater or equal to `b`.

I also added test that checks if results are in fact in `[a,b)` range (by using `range_test_engine` class that generates max random value or zeroes).

### DIFF
--- a/test/test_uniform_real_distribution.cpp
+++ b/test/test_uniform_real_distribution.cpp
@@ -14,10 +14,13 @@
 #include <boost/compute/system.hpp>
 #include <boost/compute/command_queue.hpp>
 #include <boost/compute/algorithm/count_if.hpp>
+#include <boost/compute/algorithm/transform.hpp>
+#include <boost/compute/function.hpp>
 #include <boost/compute/container/vector.hpp>
 #include <boost/compute/random/default_random_engine.hpp>
 #include <boost/compute/random/uniform_real_distribution.hpp>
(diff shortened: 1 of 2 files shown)


haahh:
Travis-CI tests failed only because of `find_end` algorithm. It's fixed by https://github.com/boostorg/compute/pull/597.

test/test_uniform_real_distribution.cpp
```diff
@@ -40,7 +43,60 @@ distribution.generate(vec.begin(), vec.end(), engine, queue);
...
+BOOST_AUTO_TEST_CASE(uniform_real_distribution_range)
+{
+    using boost::compute::lambda::_1;
+
+    boost::compute::vector<float> vec(32, context);
+
+    // initialize the range_test_engine
+    range_test_engine<boost::compute::uint_> engine(queue);
+
+    // setup the uniform distribution to produce floats between 1 and 100
```
kylelutz: Shouldn't this be between `0.9` and `1.0`?

haahh: You're right! I'll fix this in a minute.

haahh: Fixed.

PR #596, opened 2016-04-24
uniform_real_distribution: fix boundary so values are really [a,b) instead of [a,b]

jagerman:
Use the same nextafter approach from boost::compute::random::normal_distribution to eliminate the possibility of getting a draw equal to b, so that we really do produce [a,b) values, as currently documented (and desired).


haahh:
Just 15 minutes ago I pushed the same PR. Well without the assertion `BOOST_ASSERT(a < b);` but with additional test.

jagerman:
Yeah, I just noticed that.  Closing this one.

haahh:
I'll add the assert. It's a good idea.

PR #597, opened 2016-04-24
Fix find_end algorithm

haahh:
I've found out what's wrong in the `find_end` algorithm. Using uninitialized memory (from the end of `matching_indices`) was causing undefined behaviour. It was uninitialized because `search_kernel` does not initialize the last `<pattern_size> - 1` elements of `matching_indices` as the pattern can not start there. Now it's fixed.


kylelutz:
Thanks for fixing this!

PR #598, opened 2016-04-24
Increase make_literal precision

jagerman:
boost::compute::detail::make_literal is losing some precision when making literal floating point values because it uses std::numeric_limits<T>::digits10, but that only gives us the number of decimal digits that will survive a decimal->native>decimal conversion; what we are doing is a native->decimal->native conversion, which can require up to ::max_digits10 (which is larger than ::digits10 by 2 (double) or 3 (float)).

However, max_digits10 is a c++11 feature, so this commit uses digits10 + 3 when the c++11 numeric_limits isn't available.  That should be enough for an IEEE-style float, though is slightly more than necessary for some (e.g. double).

I also added a test for this (which fails with the current conversion for both float and double, and passes with the proposed commits).

### DIFF
--- a/include/boost/compute/detail/literal.hpp
+++ b/include/boost/compute/detail/literal.hpp
@@ -27,7 +27,15 @@ template<class T>
 std::string make_literal(T x)
 {
     std::stringstream s;
-    s << std::setprecision(std::numeric_limits<T>::digits10)
+    s << std::setprecision(
+#ifndef BOOST_NO_CXX11_NUMERIC_LIMITS
+            std::numeric_limits<T>::max_digits10
+#else
+            // We don't have max_digits10, so add 3 other digits (this is what is required for
+            // float, and is one more than required for double).
+            3 + std::numeric_limits<T>::digits10
+#endif
+            )
       << std::scientific
       << x;
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -225,6 +225,7 @@ add_compute_test("experimental.tabulate" test_tabulate.cpp)
 add_compute_test("misc.amd_cpp_kernel_language" test_amd_cpp_kernel_language.cpp)
 add_compute_test("misc.lambda" test_lambda.cpp)
 add_compute_test("misc.user_defined_types" test_user_defined_types.cpp)
+add_compute_test("misc.literal_conversion" test_literal_conversion.cpp)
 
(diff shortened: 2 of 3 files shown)


kylelutz:
Thanks for the fix!

haahh:
I've just noticed that on OSX it gives errors ([link](https://travis-ci.org/boostorg/compute/jobs/125438832#L2223)):

```bash
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:29: error: in "literal_conversion_float": check iss >> x has failed
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:30: error: in "literal_conversion_float": check char(iss.get()) == 'f' has failed [0xffffffff != 'f']
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:29: error: in "literal_conversion_float": check iss >> x has failed
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:30: error: in "literal_conversion_float": check char(iss.get()) == 'f' has failed [0xffffffff != 'f']
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:29: error: in "literal_conversion_float": check iss >> x has failed
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:30: error: in "literal_conversion_float": check char(iss.get()) == 'f' has failed [0xffffffff != 'f']
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:37: error: in "literal_conversion_float": check values[0] == roundtrip[0] has failed [1.23456788 != 0]
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:38: error: in "literal_conversion_float": check values[1] == roundtrip[1] has failed [1.234568 != 0]
/Users/travis/build/boostorg/compute/test/test_literal_conversion.cpp:39: error: in "literal_conversion_float": check values[2] == roundtrip[2] has failed [1.23456812 != 0]
```

jagerman:
I'm at a loss as why that is happening; my best guess is some OSX/clang++ bug w.r.t. std::istringstream operators and the stream status (I found at least one reference suggesting that, for some versions, peek() doesn't set eof as expected).

I've rewritten the test code in PR #607 -- I don't know whether that will fix it (I don't have an easily accessible OS X machine), but if it fails, it should at least help figure out which part of the conversion is failing.

PR #599, opened 2016-04-25
Turn off coveralls for POCL builds

haahh:
It takes a long time to run POCL builds and having coveralls in them only makes them fail due to exceeding the maximum time limit (see https://travis-ci.org/boostorg/compute/jobs/125418210). We will still have coveralls stats from AMD APP builds with gcc (OpenCL 1.2 and 2.0).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -410,6 +410,9 @@ script:
       fi
 
 after_success:
-    - lcov --directory test --base-directory ../include/boost/compute/ --capture --output-file coverage.info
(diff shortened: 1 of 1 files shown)


haahh:
Ah.. It still sometimes takes too much time... 

We can consider not running tests in POCL+gcc builds (clang builds are OK) until POCL becomes faster or just ignore it and wait for the next release of POCL which should be faster (POCL master branch is now faster than current release [0.13] or at least was because we did not have those problems before switching from master to 0.13).

PR #603, opened 2016-04-30
Fix sign comparison warning in test_literal_conversion

kylelutz:


haahh:
:+1:  I was about to do the same.

PR #607, opened 2016-05-02
Literal test simplification

jagerman:
The tests are failing under OS X/clang, but it isn't obvious to me why, or exactly which part is failing.  This commit simplifies the test code by ignoring the stringstream I/O status, and reorganizes it to be clearer whether it's the initial conversion to string that is failing, or the subsequent conversion back from string.

(This may or may not clear up the OSX/clang failure, but it's better test code anyway).

This also incorporates the PR #603 fix.

### DIFF
--- a/test/test_literal_conversion.cpp
+++ b/test/test_literal_conversion.cpp
@@ -29,13 +29,13 @@ BOOST_AUTO_TEST_CASE(literal_conversion_float)
         // 6 characters (5 digits) look good
         BOOST_CHECK_EQUAL(literal.substr(0, 6), "1.2345");
 
-        // Stuff the literal into a stream then extract it, and make sure we get a numerically
-        // identical result:
-        std::istringstream iss(literal);
+        // libc++ stream extraction fails on a value such as "1.2f" rather than extracting 1.2
+        // and leaving the f; so check the "f", then slice it off for the extraction below:
+        BOOST_CHECK_EQUAL(literal.substr(literal.length()-1), "f");
+
+        std::istringstream iss(literal.substr(0, literal.length()-1));
         float x;
-        std::string suffix;
-        iss >> x >> suffix;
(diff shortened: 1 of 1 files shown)


jagerman:
This did not fix the OSX/clang error ([link](https://travis-ci.org/boostorg/compute/jobs/127370548#L1769)), and I'm not sure why.

The conversion to float is going okay, because the first test:

```BOOST_CHECK_EQUAL(literal.substr(0, 6), "1.2345");```

is not failing--so we have something starting with 1.2345, as desired, and it's probably the right value.  But then, from the test output, this code:

```
        std::istringstream iss(literal);
        float x;
        std::string suffix;
        iss >> x >> suffix;
```

seems to be completely broken: we end up with x == 0.f, and suffix (which should be "f") is empty.  Am I doing something wrong in the std::istringstream use here?

jagerman:
[The bug is in libc++](https://llvm.org/bugs/show_bug.cgi?id=17782), and shows up with the clang++/libc++ combination on linux, too: libc++ fails if asked to extract float/double values followed various special letters, of which "f" is one.

The cce140e commit I just pushed works around the problem, which should resolve the test failure.

haahh:
OK, I also did some testing after I'd noticed those errors and it looked like there's a bug in apple clang/libc++ but somehow I couldn't find reference to the bug. Thanks!

Fortunately, this bug should not effect Boost.Compute. https://github.com/boostorg/compute/pull/603 was merged so you can rebase your branch on the `develop` if you prefer.

jagerman:
Just rebased the branch onto develop.

Indeed the bug is only in the test program, not the actual literal generation, but eliminating the false positive test failure under clang/libc++ (not just on OS X) seems worthwhile, if minor.

kylelutz:
Merged! Thanks!

PR #610, opened 2016-05-05
OSX

haahh:
This PR resolves issues with Boost.Compute running on OSX.

Tests for `detail::inplace_reduce`, `radix_sort`, `radix_sort_by_key` and matrix transpose example are disabled for CPU devices on Apple platform because for CPU devices on Apple platform when local memory is used in a kernel local work group size must be [1;1;1]. Those algorithms and matrix example are designed for GPUs, so from practical perspective we do not lose any functionality.

I've also fixed `boost::compute::discrete_distribution` constructor. There was a read from uninitialized memory `m_probabilities[i-1]` (with `i = 0`).

I've found out that there's a bug in `clCreateBuffer()` on Apple platform. `clCreatBuffer()` does not return `NULL` and does not set error to `CL_INVALID_BUFFER_SIZE` when size of the buffer memory object is greater than `CL_DEVICE_MAX_MEM_ALLOC_SIZE`. You can see bug example [here](https://travis-ci.org/haahh/compute/builds/128017494#L1926) - 42949672960 is the size of allocated memory where `CL_DEVICE_MAX_MEM_ALLOC_SIZE` is only 1073741824 bytes.

After this PR, Travis-CI OSX build is not allowed to fail.

### DIFF
--- a/include/boost/compute/random/discrete_distribution.hpp
+++ b/include/boost/compute/random/discrete_distribution.hpp
@@ -11,6 +11,9 @@
 #ifndef BOOST_COMPUTE_RANDOM_DISCRETE_DISTRIBUTION_HPP
 #define BOOST_COMPUTE_RANDOM_DISCRETE_DISTRIBUTION_HPP
 
+#include <boost/type_traits.hpp>
+#include <boost/static_assert.hpp>
+
 #include <boost/compute/command_queue.hpp>
 #include <boost/compute/function.hpp>
 #include <boost/compute/algorithm/accumulate.hpp>
@@ -114,6 +117,11 @@ public:
 private:
     size_t m_n;
     ::std::vector<double> m_probabilities;
+
+    BOOST_STATIC_ASSERT_MSG(
+        boost::is_integral<IntType>::value,
+        "Template argument must be integral"
+    );
 };
 
 } // end compute namespace
--- a/include/boost/compute/random/discrete_distribution.hpp
+++ b/include/boost/compute/random/discrete_distribution.hpp
@@ -42,8 +42,8 @@ public:
     /// the range [\p first, \p last)
     template<class InputIterator>
     discrete_distribution(InputIterator first, InputIterator last)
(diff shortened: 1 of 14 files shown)


include/boost/compute/random/discrete_distribution.hpp
```diff
@@ -42,8 +45,8 @@ class discrete_distribution
     /// the range [\p first, \p last)
     template<class InputIterator>
     discrete_distribution(InputIterator first, InputIterator last)
-        : m_n(std::distance(first, last)),
-          m_probabilities(std::distance(first, last))
+        : m_n((std::max)(size_t(1), static_cast<size_t>(std::distance(first, last)))),
+          m_probabilities((std::max)(size_t(1), static_cast<size_t>(std::distance(first, last))))
```
kylelutz: I think we still need to handle the case where `first == last` better. Looking at the documentation for [std::discrete_distribution](http://en.cppreference.com/w/cpp/numeric/random/discrete_distribution/discrete_distribution), when the input range is empty (or calling the default constructor), we should just initialize with a single probability of `1`. But this is currently not implemented, and we don't have to do it in this pull request. I'll open a separate bug for this.

include/boost/compute/random/discrete_distribution.hpp
```diff
@@ -52,9 +55,14 @@ class discrete_distribution
             sum += *iter;
         }
 
-        for(size_t i=0; i<m_n; i++)
+        InputIterator iter = first;
+        m_probabilities[0] = (*iter)/sum;
```
kylelutz: Yeah, this could be invalid if `first == last`.

kylelutz:
Thanks for all the fixes!

include/boost/compute/random/discrete_distribution.hpp
haahh: Oh, yeah, you're right. I'll fix it.

Issue #612, opened 2016-05-10
Add default constructor for discrete_distribution


kylelutz:
Add a default constructor to [`discrete_distribution`](http://www.boost.org/doc/libs/develop/libs/compute/doc/html/boost/compute/discrete_distribution.html). Also support the case `first == last` when calling the current iterator range constructor. See the documentation for [`std::discrete_distribution`](http://en.cppreference.com/w/cpp/numeric/random/discrete_distribution/discrete_distribution).

haahh:
Can be closed.

PR #613, opened 2016-05-10
Default constructor for discrete distribution

haahh:
Resolves https://github.com/boostorg/compute/issues/612. I've also added `min()` and `max()` methods and changed `boost::compute::discrete_distribution` so now it generates `IntType` values instead of always generating `uint` values.

### DIFF
--- a/include/boost/compute/random/discrete_distribution.hpp
+++ b/include/boost/compute/random/discrete_distribution.hpp
@@ -94,6 +94,24 @@ public:
         return m_probabilities;
     }
 
+    /// Returns the minimum potentially generated value.
+    result_type min BOOST_PREVENT_MACRO_SUBSTITUTION () const
+    {
+        return result_type(0);
+    }
+
+    /// Returns the maximum potentially generated value.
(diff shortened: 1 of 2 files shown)


include/boost/compute/random/discrete_distribution.hpp
```diff
@@ -71,52 +88,65 @@ class discrete_distribution
...
-
     /// Returns the probabilities
     ::std::vector<double> probabilities() const
     {
         return m_probabilities;
     }
 
-    /// Generates uniformily distributed integers and stores
+    /// Returns the minimum potentially generated value.
+    result_type BOOST_PREVENT_MACRO_SUBSTITUTION min() const
```
kylelutz: I think the `BOOST_PREVENT_MACRO_SUBSTITUTION` macro needs to go between the function name and the parentheses. Something like:

```
result_type min BOOST_PREVENT_MACRO_SUBSTITUTION () const
```

haahh: Thanks! Fixed!

kylelutz:
Looks good, merged!

Issue #363, opened 2014-12-29
Fix type-safety issues when copying to/from host-memory


kylelutz:
There are cases where Boost.Compute blindly copies raw bytes of memory from the device to the host which are then interpreted wrong (e.g. 4-byte `float` on the device to 8-byte `double` on the host yields garbage values). Boost.Compute should automatically convert the values to the proper type when copying.

Example from the Boost.Compute review:
```
compute::device device = compute::system::default_device();
compute::context context(device);
compute::command_queue queue(context, device);

// generate random data on the host - type is float
std::vector<float> host_vector(10000);
std::generate(host_vector.begin(), host_vector.end(), rand);

// create a vector on the device
compute::vector<float> device_vector(host_vector.size(), context);

// transfer data from the host to the device
compute::copy(host_vector.begin(), host_vector.end(), device_vector.begin(), queue);

double reduction_result = 0.0; // result is of type double
compute::reduce(device_vector.begin(), device_vector.end(), &reduction_result, queue);
std::cout << "result: " << reduction_result<< std::endl;
```

Another example:
```
// generate random data on the host - type is float
std::vector<float> host_vector(10000);
std::generate(host_vector.begin(), host_vector.end(), rand);

// create a vector on the device - type is double
compute::vector<double> device_vector(host_vector.size(), context);

// transfer data from the host to the device
compute::copy(host_vector.begin(), host_vector.end(), device_vector.begin(), queue);
```

haahh:
Currently, I'm working on copying data from host to the device and I have some questions.

```cpp
template<class InputIterator, class OutputIterator>
inline OutputIterator copy(InputIterator first,
                           InputIterator last,
                           OutputIterator result,
                           command_queue &queue = system::default_queue())
{
    (...)
}
```

I have 3 implementations: 
 1. Copying data `[first; last)` on host to `std::vector<typename OutputIterator::value_type>` and then just performing `queue.enqueue_write_buffer(..)` to the `result`.
 2. Mapping `result` into host using `queue.enqueue_map_buffer(...)`, performing `std;:copy` (since `result` is mapped to host) and at the end unmaping `result` ([code](https://gist.github.com/haahh/5605c1057f6f0282437c777dd33ea24f)).
 3. Doing this (or something like this, I'm not sure if it works with structures):
```cpp
boost::compute::mapped_view<int> mapped_host(host_vector.data(), host_vector.size(), context);
boost::compute::transform(mapped_host.begin(),
                          mapped_host.end(),
[...]

PR #616, opened 2016-05-26
Type-safe copying

haahh:
Resolves https://github.com/boostorg/compute/issues/363.

#### Sync. copying to/from device

For both coping data to and from device there are 3 `dispatch_copy()` functions and each function works in different scenario:

1. Copying to device when value types (of `InputIterator` and `OutputIterator`) match and host iterator is contiguous
2. Copying to device when value types mismatch and host iterator is contiguous
3. Copying to device when host iterator is not contiguous (values types does not have to match)

##### 1st option
It's straightforward copying using OpenCL functions for reading and writing a device memory.

##### 2nd option has three sub-options
1. Maping device memory to host and performing `std::copy` (casting/converting and copying on host)
2. Using `std::vector<DeviceIterator::value_type>` as a intermediate memory  (casting/converting on host, copying using OpenCL function)
3. Maping host memory to the device and using copy kernel to both convert and copy the data.

##### 3rd option has two sub-options
1. Maping device memory to host and performing `std::copy` (casting/converting and copying on host)
2. Using `std::vector<DeviceIterator::value_type>` as a intermediate memory  (casting/converting on host, copying using OpenCL function)

##### Additional info
When copying to host memory we treat host iterators whose value type is `bool` as if they were non-contiguous.

#### Async. copying to/from device

There is asynchronous copying with mismatching types works too. In both cases (to/from device) I used maping host memory to the device and using copy kernel to both convert and copy the data.

####  Various fixes

* Fix types in `perf_sort_by_key.cpp`
* Fix return values in async coping `svm_ptr<>` from/to device
* Fix test in test_functional_as.cpp


example/opencv_histogram.cpp
```diff
@@ -11,7 +11,9 @@
 //Code sample for calculating histogram using OpenCL and
 //displaying image histogram in OpenCV.
 
-#define BOOST_COMPUTE_DEBUG_KERNEL_COMPILATION
+#ifndef BOOST_COMPUTE_DEBUG_KERNEL_COMPILATION
+    #define BOOST_COMPUTE_DEBUG_KERNEL_COMPILATION
+#endif
```
kylelutz: We should probably just remove this define from the file. It should only really be defined by the user when they are debugging.

haahh: Will fix that later today.

include/boost/compute/algorithm/copy.hpp
```diff
@@ -114,29 +307,266 @@ dispatch_copy_async(InputIterator first,
...
+            ::boost::addressof(*result)
+        )
+    );
+    copy_on_device(
+        first,
+        last,
+        make_buffer_iterator<output_type>(mapped_host),
+        queue
+    );
+    // update host memory asynchronously by maping and unmaping memory
```
kylelutz: Shouldn't `asynchronously` here be `synchronously`? Also, I think an alternative would be instead of mapping and unmapping the buffer, replace the call above with `copy_on_device_async()` and wait on the returned event/future.

haahh: Yes, indeed, it should be `synchronously` since later `wait()` is called for event returned by unmap operation. 

And yes, those lines can be replace by just calling `copy_on_device_async()` as it does the same (create buffer for mapping host memory, perform copying on device and map/unmap buffer).

include/boost/compute/algorithm/detail/copy_to_device.hpp
```diff
@@ -53,6 +53,49 @@ inline DeviceIterator copy_to_device(HostIterator first,
...
+            offset * sizeof(value_type),
+            count * sizeof(value_type)
+        )
+    );
+
+    // copy [first; last) to result buffer
+    std::copy(first, last, pointer);
+
+    // unmap result buffer
+    boost::compute::event unmapEvent = queue.enqueue_unmap_buffer(
```
kylelutz: Can you change this to `unmap_event` to match the Boost style?

test/test_functional_as.cpp
```diff
@@ -52,7 +52,7 @@ BOOST_AUTO_TEST_CASE(roundtrip_int_float)
 
     // check values
     CHECK_RANGE_EQUAL(
-        int, 8, output, (1, 2, 3, 4, 5, 6, 7, 8)
+        int, 8, input, (1, 2, 3, 4, 5, 6, 7, 8)
```
kylelutz: Good catch :-).

kylelutz:
Looks good! I'll finish up reviewing/testing this out today.

haahh:
Just fyi, I ran all the tests with SVM on Windows with the latest AMD drivers with fixed `clEnqueueSVMMemcpy()` (you can see that there are tests with `svm_ptr<>` that are skipped for AMD, this is the reason).

include/boost/compute/algorithm/copy.hpp
haahh: Fix: `dispatch_copy_async()`, not `copy_on_device_async()`

haahh:
I've just found that that there is an inconsistency in what async. copy functions return when `first == last`.  In one function it's future object with `result` iterator and an empty event (`event()`), in other functions it's empty future object... However, from user perspective it does not matter because it's impossible to get the value in both of those cases (user gets `Invalid event` error or sth like this when try to `get()` the value).

kylelutz:
Awesome! Merged!

PR #617, opened 2016-06-02
Add Travis CI build with latest Boost libs

haahh:
Two additional Travis-CI builds (gcc and clang) to be sure that Boost.Compute works with libraries from the latest Boost version. Spoiler: it works.

POCL builds may not work now. It has nothing to do with this pull-request, it's just because llvm [turned off their APT mirror](http://lists.llvm.org/pipermail/llvm-foundation/2016-May/000020.html) and gpg key can't be download.


kylelutz:
Looks good, merged!

PR #620, opened 2016-06-23
Temp. fix for llvm apt being down

haahh:
Apt for llvm and clang has been down for quite some time (see https://github.com/travis-ci/travis-ci/issues/6120), so here is the fix for POCL builds which require the latest clang and llvm.


kylelutz:
Thanks!

PR #622, opened 2016-07-01
Fix adjacent_difference when first == result

haahh:
Situation where `first == result` was only covered when binary function was not specified.


kylelutz:
Looks good! We should probably also add a test for the `first == last` case.

PR #629, opened 2016-07-10
Add missing include for std::cerr (on macOS)

thtrummer:


kylelutz:
Thanks for the patch!

PR #631, opened 2016-07-10
Fix build error due to possible typo (on macOS)

thtrummer:


haahh:
Yes, indeed. Those are typos. This `uint` typedef'ed in Linux is so annoying.

kylelutz:
Thank for finding/fixing this!

PR #632, opened 2016-07-16
Various fixes

haahh:
This fixes 3 things:

* `boost::compute::array<T>` was not working correctly with context different than `system::default_context()` since it was using `default_queue()` in its methods (so wild `Invalid context` errors could appear).
* I added

```cpp
template<class OtherAlloc>
vector& operator=(const vector<T, OtherAlloc> &other)
{
}
```
for `boost::compute::vector<T, A>, so now something like this:

```cpp
vector<int, some_allocator> b(context);
vector<int, different_allocator> a(context);
vector<int, some_allocator> c = a; // this works in `develop`
a = b; // this does not work in `develop`, works now
```

works.

* I changed several tests so in the future we can change `Context::context, Context::queue, Context::device` and tests will run using those context, queue and device, not the default context, queue and device.

In October I will work more on Boost.Compute tests and try to somehow organize them and also increase coverage.

### DIFF
--- a/include/boost/compute/container/array.hpp
+++ b/include/boost/compute/container/array.hpp
@@ -73,20 +73,33 @@ public:
     array(const array<T, N> &other)
         : m_buffer(other.m_buffer.get_context(), sizeof(T) * N)
     {
-        boost::compute::copy(other.begin(), other.end(), begin());
+        command_queue queue = default_queue();
+        boost::compute::copy(other.begin(), other.end(), begin(), queue);
+        queue.finish();
     }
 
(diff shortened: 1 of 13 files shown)


haahh:
I added one more commit. https://github.com/boostorg/compute/pull/632/commits/4c21633bf9351afdd2e4d201dbcabda15ced0b61 fixes problem with copy ctor of `boost::compute::vector`which is causing `Invalid Context` error when context of the vector that is being copied is not default context.

kylelutz:
Looks good, thanks for all the fixes!

PR #636, opened 2016-07-27
Fixed bug in opengl_create_shared_context

dPavelDev:
Sorry for the delay, I forgot how to rollback commits :)


kylelutz:
Good catch! Thanks for the patch!

Issue #637, opened 2016-07-28
Compile warnings in master and develop branches


dPavelDev:
When I tried to rebuild my project, I got some compile warnings in boost::compute sources:

In master and develop branches:
```
e:\PROJECTS\compute\include\boost\compute\exception\context_error.hpp:73: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
     const size_t get_private_info_size() const throw()
                                                      ^
```
If remove const before size_t, warning is disappear but may be exist more right method.

In develop branch:
```
e:\PROJECTS\compute\include\boost\compute\kernel.hpp:265: warning: unused parameter 'index' [-Wunused-parameter]
     void set_arg_svm_ptr(size_t index, void* ptr)
                                 ^
```
And ptr when not defined CL_VERSION_2_0.

Also, when I rebuild project using MSVC (argument size_t size):
```
E:\PROJECTS\compute\include\boost\compute\detail\get_object_info.hpp:42: warning: C4267: 'argument': conversion from 'size_t' to 'cl_uint', possible loss of data
```
acquire.hpp:39, acquire.hpp:64, get_object_info.hpp:42
```
E:\PROJECTS\compute\include\boost\compute\interop\opengl\acquire.hpp:39: warning: C4267: 'argument': conversion from 'size_t' to 'cl_uint', possible loss of data
```
On x64 size_t != cl_uint. It is necessary to fix this warnings?

kylelutz:
Most of these should be fixed with PR #638. Would you mind testing that out?

dPavelDev:
Good job! Only 1 warning:
```
E:\PROJECTS\boostorg\compute\include\boost\compute\detail\get_object_info.hpp:42: warning: C4267: 'argument': conversion from 'size_t' to 'cl_uint', possible loss of data
```

kylelutz:
Hmm, is there any sort of "back-trace" for that warning? I'd like to see where it's getting called from (the `get_object_info()` function is a wrapper for many different `clGet*Info()` calls, most of which use `size_t` instead of `cl_uint`).

haahh:
I've found this. `kernel.hpp`, lines 191 and 268, `index` has to be cast to `cl_uint`, and there's no warnings after that.

kylelutz:
Ahh, got it. This should be fixed in PR #638 now.

PR #638, opened 2016-07-31
Fix warnings

kylelutz:


kylelutz:
This should fix most of the warnings reported in issue #637.

kylelutz:
This should fix all of your warnings now.

PR #641, opened 2016-08-14
Fixed unnecessary reallocation in vector::resize() when size == capacity

dPavelDev:


haahh:
It think there is one commit that's not needed.

dPavelDev:
Ops... I don't understand how it happened.

haahh:
You don't need to close the pull request, you can just do a forced push to the branched used for this pull request.

PR #644, opened 2016-08-16
Distributed work infrastructure (context, queue, vector)

haahh:
This is a "beta" version of distributed work infrastructure: `distributed::context`, `distributed::command_queue`, `distributed::vector`. I'm open to comments and suggestions.

The idea is pretty simple: `distributed::context` is a wrapper for many OpenCL contexts, `distributed::command_queue` - for many OpenCL command queues, and `distributed::vector` is a container that allocates memory across many devices according to certain weight function.

### Ongoing work

Currently, I'm working on `distributed::vector` and `copy` algorithm (for now it's just a set of `vector`'s private methods, but it'll change soon).


haahh:
I think about removing `blocking` argument from `distributed::vector` ctors and methods and instead use `boost::optional` for an optional reference to a `wait_list` object, but that would add additional dependency. Maybe I'll just remove it for first release and later think about this.

I also think I'll remove the queue from `distributed::vector`. Then I'll also have to change some of its constructors so they don't accept `distributed::context` but `distributed::command_queue` (empty ctor and ctor creating vector with count copies of some value).

haahh:
I added `distributed::copy()` algorithm with tests. For now `distributed::copy()` has many different lists of arguments depending on what your copying and in which way. After I introduce `distributed_iterator()`, it will be slightly simpler.

haahh:
Also, there are commits which are present in https://github.com/boostorg/compute/pull/648, but I'll remove them before this pull request is ready to be merged. @kylelutz any comments on code?

haahh:
In `copy_async()` I can't use `wait_list` since it does not support event from different contexts. For example `wait()` method does not work properly when wait_list contains events from different contexts, which can happen using `distributed::vector`.

In 2nd half of September I will introduce changes to `wait_list` so it force user to store only `event`s from the same OpenCL context, and maybe make new class for wrapping many events from different contexts (based on `future` or `wait_list`).

haahh:
`distributed::transform()` should have good performance, it's indirectly influenced by the weight function (currently all weights are equal, I will later add weight function based on device performance) since it waits for the slowest device.

`distributed::reduce()` should have good performance when all used devices (queues) are in the same context, but I have an idea how to make it works with the same speed in other cases.

PR #648, opened 2016-08-22
Misc fixes

haahh:
This fixes:

* Some typos 
* `C4267` warning when compiling with MSVC and OpenCL 2.0
* `gather()` algorithm, `m_offset` was unnecessary and caused errors since offset is already stored in an iterator.


kylelutz:
Looks good! Will try to get this into master before 1.62.

PR #649, opened 2016-08-23
Added accessors to vector types

amirshavit:
This basically replaces the protected m_values member with public x/y, x/y/z/w, s0..s7, s0..sf members according to the vector size.

### DIFF
--- a/include/boost/compute/types/fundamental.hpp
+++ b/include/boost/compute/types/fundamental.hpp
@@ -109,7 +109,7 @@ public:
 
     vector_type()
     {
-
+        BOOST_STATIC_ASSERT(sizeof(Scalar) * N == sizeof(vector_type<Scalar, N>));
     }
 
     explicit vector_type(const Scalar scalar)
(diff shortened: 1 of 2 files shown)


haahh:
Is this approach safe? I mean, if I recall correctly a compiler can introduce a padding between two member variables. Of course I don't think that would happen in that case when using any popular C++ compiler for x86 like g++ or clang++, but it's still a question.

amirshavit:
I don't think it's less safe than an array of scalars, but I'll try to find some text to support that.
Either way, a static assert of ```sizeof(Scalar) * N == sizeof(vector_type<Scalar, N>)``` can remove all concerns.

haahh:
> I don't think it's less safe than an array of scalars, but I'll try to find some text to support that.

I am positive that a compiler can't put padding into an array of scalars.

> Either way, a static assert of sizeof(Scalar) * N == sizeof(vector_type<Scalar, N>) can remove all concerns.

:+1:

jszuppe:
Done in #823

Issue #647, opened 2016-08-22
Compile using (/for) an older OpenCL API version


amirshavit:
Currently, if the cl.h header supports newer API versions (say CL_VERSION_2_1), boost::compute uncontrollably uses that API version. If I run my compiled binary on a different system, with an older opencl library, the binary will fail to load due to missing exports in the dll/so.

I see there are run-time version checks (`device.check_version(...)`) but they aren't helpful for this problem.

An easy solution could be if the cl.hpp (of boost::compute) could #undef those (after `#include <.../cl.h>`), something like -

``` c++
#if defined(BOOST_COMPUTE_AVOID_CL_VERSION_2_0) && defined(CL_VERSION_2_0)
#undef CL_VERSION_2_0
#endif
...
```
##

haahh:
Yeah, IIRC that was mentioned a few times. Basically, you postulate to add macros for selecting the newest supported OpenCL version instead of using the newest defined in used `cl.h`, yes?

amirshavit:
Yes.

An alternative form (based on your suggestion) -
```c++
#if defined(BOOST_COMPUTE_MAX_CL_VERSION)
// ...
#  if defined(BOOST_COMPUTE_MAX_CL_VERSION < 20) && defined(CL_VERSION_2_0)
#    undef CL_VERSION_2_0
#  endif
// ...
#endif
```

I'm not sure if `#undef`-ing is the so-called right solution, but it sure it the fastest.

haahh:
I think the right way would be to define internal macros based on OpenCL version macros and options set by a user, and replace OpenCL version marcos (like `CL_VERSION_2_0`) with them.

amirshavit:
I agree.  Should I open a pull request or is this something for further consideration?

haahh:
If you want to do this, then do this and open a pull request and wait for @kylelutz to accept it. If you don't have time for this, then you have to wait for me or @kylelutz to code this, but I don't have time to do that this month.

haahh:
Fixed - https://github.com/boostorg/compute/pull/699.

PR #650, opened 2016-08-23
Added BOOST_COMPUTE_MAX_CL_VERSION to specify the newest version to use

amirshavit:
See #647

### DIFF
--- a/include/boost/compute/algorithm/detail/copy_on_device.hpp
+++ b/include/boost/compute/algorithm/detail/copy_on_device.hpp
@@ -144,7 +144,7 @@ inline future<OutputIterator> copy_on_device_async(InputIterator first,
     return make_future(result + std::distance(first, last), event_);
 }
 
-#ifdef CL_VERSION_2_0
+#ifdef BOOST_COMPUTE_CL_VERSION_2_0
 // copy_on_device() specialization for svm_ptr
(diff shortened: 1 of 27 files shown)


kylelutz:
Thanks for the patch! This does seem like a good approach, though with a big change like this we'll need to do some testing. Will try to get this merged in the next Boost release.

haahh:
I'll test it on 1.2, 2.0 and 2.1 implementations, check if it covers all '#ifdef`s.

haahh:
You did not apply changes (`CL_VERSION_X_Y` to `BOOST_COMPUTE_CL_VERSION_X_Y`) in tests. 

Also, I think that in each
```cpp
#if defined(CL_VERSION_X_Y)
#  if !defined(BOOST_COMPUTE_MAX_CL_VERSION) || BOOST_COMPUTE_MAX_CL_VERSION >= X0Y
#    define BOOST_COMPUTE_CL_VERSION_X_Y
#    ifndef CL_USE_DEPRECATED_OPENCL_X_Y_APIS // new
#      define CL_USE_DEPRECATED_OPENCL_X_Y_APIS // new
#    endif // new
#  endif
#endif
```
we should define macro `CL_USE_DEPRECATED_OPENCL_X_Y_APIS`.

PR #651, opened 2016-08-23
Fix minmax_element() and add test for it

haahh:
Remove useless `Compare` template parameter from `minmax_element()` overload, which was causing errors, and add test for this function.


kylelutz:
Good catch! Merged!

haahh:
Thanks!

PR #654, opened 2016-08-28
Document rough space complexity for every algorithm

haahh:
This documents space complexity for every algorithm. Some are exact values, some are  rounded up estimations.


kylelutz:
Merged! Thanks for taking the time to do all this!

PR #657, opened 2016-09-14
check result of flush and finish

Slonegg:
At least on NVIDIA most of the OpenCL commands like clEnqueueNDRangeKernel won't return an error if it happened inside kernel. Instead failure will be reported when command queue is flushed. So it is convenient to check errors in kernel by adding clFinish call after enqueue.


kylelutz:
Thanks for the patch! Will get this merged in after the upcoming 1.62 release (should be in the next week or so).

kylelutz:
Merged! Thanks!

PR #658, opened 2016-09-14
added headers to program::compile

Slonegg:


Slonegg:
@kylelutz do you think we can add support for OpenCL headers to boost::compute? They're convenient if you include them in multiple files. Caching would be nice too...

haahh:
>Caching would be nice too...

I think for caching result of `clCompileProgram(..)` you would need to store program source hash in `boost::compute::program` class.

keryell:
Yes.
But is this an issue? The hash is only 160 bits.

But more difficult seems to figure out what is really included by the compiler itself, which is outside of the Boost.Compute realm...
This will depend on the options passed to the compiler that we need to understand, such as the `-I` paths and so on...
So the best place to do this kind of caching is inside the OpenCL stack itself, such as done by PoCL or nVidia.

Slonegg:
I guess including all headers, compilations options and source of the program in hash should be more or less robust solution. Maybe warn user if he uses -I option, because it may trick caching system.

haahh:
>This will depend on the options passed to the compiler that we need to understand, such as the -I paths and so on...
So the best place to do this kind of caching is inside the OpenCL stack itself, such as done by PoCL or nVidia.

We already do offline-caching of OpenCL programs... so I imagine even right now you can pass `-I` option to build options and shoot yourself in foot by changing file/s you included. I think he only way to make Boost.Compute caching 100% robust would be to base hash not on program source, but on IL or binary. Unfortunately, binaries or IL are usually much bigger.

Anyway, MR looks ok. I'll merge it. Thanks!

Issue #676, opened 2016-12-20
Unexpected context is created by opengl_create_shared_context()


sonogi1:
Computer: Intel CPU with HD Graphics.
OS: Windows 10
SDK: Intel SDK for OpenCL™ Applications 2016 R3

Issue: 
The SDK includes two platforms, "Experimental OpenCL 2.1 CPU Only" and "Intel(R) OpenCL". The former CPU only platform is unexpectedly selected by opengl_create_shared_context. opengl_create_shared_context(platform const&) may be needed for manual platform selection.

kenba:
Computer: AMD A10 with 6 Graphics cores and an Nvidia GeForce GTX 750 Ti Graphics Card
OS: Windows 10
SDKs: AMD APP SDK 3.0 and Intel SDK for OpenCL™ Applications 2016 R3

Issue:
The computer has multiple platforms, most of which support OpenGL... 
When calling opengl_create_shared_context() the program crashed during the call to "GetGLContextInfoKHR".

According to [this](https://software.intel.com/en-us/articles/sharing-surfaces-between-opencl-and-opengl-43-on-intel-processor-graphics-using-implicit) article on Intel's site: "Careful reading of the specification highlights this is a platform extension not a device extension"...

Simply adding the following lines fixed the crash on the AMD/Nvidia system and  enabled OpenGL/OpenCL sharing (see pull request #677):

        if (!platform.supports_extension(cl_gl_sharing_extension))
          continue;

Note: opengl_create_shared_context() worked fine on a more "benign" system: an Intel Core i7-6500u  with an AMD R5 M335 graphics card also with both AMD and Intel SDKs and it still works with the fix above.

PR #677, opened 2016-12-20
Check platform extension for issue #676

kenba:
Check that a platform supports OpenCL/OpenGL sharing before attempting to load the clGetGLContextInfoKHR() extension function in function opengl_create_shared_context().


haahh:
I think OSX build keeps failing only because some bug in image used by Travis CI. It worked ok when I rebased it and pushed to my fork https://travis-ci.org/haahh/compute/builds/214949240.

PR #678, opened 2016-12-23
Add Wrapper to clfft clblas ...

mahdimcheik:
Hello,
The problem with OpenCL is that there is no complete library compared to cuda where all libraries like cufft and cublas are installed by default,  but with OpenCL  we have to search equivalent libraries and try to figure out a way to link them with other libraries we may use in the same program. I think that compute boost is a good apportunity  to create a full library cuda-like. Libraries like Arrayfire are not standard and dont give full control as in native OpenCL c/c++ program without lose of performance.
So can please you add wrapper "boost-like" of those libraries clfft clblas clsparse , clrand and maybe cl deep learning if that exist, where the programming is a kind unified under the boost model.
sorry for my english and thank you if you still red till here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -366,6 +366,7 @@ before_install:
       # OSX
       elif [[ ${TRAVIS_OS_NAME} == "osx" ]]; then
         brew update
+        /usr/bin/yes | pip uninstall numpy
         brew outdated boost || brew upgrade boost
         brew outdated cmake || brew upgrade cmake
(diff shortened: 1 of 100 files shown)


kylelutz:
Hi @mahdimcheik,

Looks like this is a feature request, would you mind closing this and opening an issue with the "feature" tag on the [issue tracker](https://github.com/boostorg/compute/issues).

Thanks!

Issue #683, opened 2017-01-13
Lambda kernel compilation failed due to ambiguous clamp call


donniet:
The following example compiles but fails at run time with the error below.

```
#include <boost/compute.hpp>
#include <iostream>

namespace compute = boost::compute;

int main() {
  // get default device and setup context
  compute::device device = compute::system::default_device();
  compute::context context(device);
  compute::command_queue queue(context, device);

  // create data array on host
  unsigned int host_data[] = { 1, 3, 5, 7, 9 };

  // create vector on device
  compute::vector<unsigned int> device_vector(5, context);

  // copy from host to device
  compute::copy(
    host_data, host_data + 5, device_vector.begin(), queue
  );

  using compute::lambda::clamp;
  using compute::lambda::_1;
  compute::uint_ zero = 0, five = 5;

  compute::transform(
    device_vector.begin(), device_vector.end(), device_vector.begin(), clamp(_1, zero, five), queue
  );

  // create vector on host
  std::vector<unsigned int> host_vector(5);

  // copy data back to host
  compute::copy(
    device_vector.begin(), device_vector.end(), host_vector.begin(), queue
  );

  return 0;
}
```

I compiled this on a Mac with this command:  
`$ clang++ -framework OpenCL -DBOOST_COMPUTE_DEBUG_KERNEL_COMPILATION main.cpp`

And when running I get the following error:
```
$ ./a.out 
Boost.Compute: kernel compilation failed (-11)
--- source ---
#define boost_pair_type(t1, t2) _pair_ ## t1 ## _ ## t2 ## _t
#define boost_pair_get(x, n) (n == 0 ? x.first ## x.second)
#define boost_make_pair(t1, x, t2, y) (boost_pair_type(t1, t2)) { x, y }
#define boost_tuple_get(x, n) (x.v ## n)


__kernel void copy(__global uint* _buf0, const uint count)
{
uint index = get_local_id(0) + (512 * get_group_id(0));
for(uint i = 0; i < 4; i++){
    if(index < count){
_buf0[index]=clamp(_buf0[index], 0, 5);
       index += 128;
    }
}

}

--- build log ---
<program source>:12:14: error: call to 'clamp' is ambiguous
_buf0[index]=clamp(_buf0[index], 0, 5);
             ^~~~~
/System/Library/Frameworks/OpenCL.framework/Versions/A/lib/clang/3.2/include/cl_kernel.h:4867:18: note: candidate function
[...]

PR #684, opened 2017-01-13
potential fix for #683

donniet:
Adding the stream operators which include the correct numeric literal suffixes for uint, long and ulong in include/boost/compute/detail/meta_kernel.hpp fixes issue #683 in my tests.  This is the same solve as for floats.


donniet:
testing failed-- more wore has to be done.

PR #689, opened 2017-02-01
Check opencl availability on test.

Flast:
This change avoids massive of test failure that's caused by no opencl env.


kylelutz:
Merged! Thanks @Flast !

PR #693, opened 2017-03-17
Various fixes

haahh:
* Fixes https://github.com/boostorg/compute/issues/683 - variable type is kept when using it in `boost::compute::lambda`. I decided that for now it's better not to change `operator<<` operators for `detail::meta_kernel` as was proposed in https://github.com/boostorg/compute/pull/684. Instead, I specialised `void operator()(proto::tag::terminal, const T &x)` in `lambda::context`.
* Remove race conditions from test in `test_async_wait_guard.cpp` and from test `destructor_callback` in `test_buffer.hpp`


kylelutz:
Awesome! Thanks for fixing these!

PR #696, opened 2017-03-24
AMD SVM bug was fixed on Windows

haahh:
`clEnqueueSVMMemcpy()` works for AMD devices on Windows (Linux Catalyst drivers still without an update).


kylelutz:
Thanks!

PR #697, opened 2017-03-24
fix SHA1 hash in build_with_source()

huanzhang12:
In build_with_source(), when BOOST_COMPUTE_USE_OFFLINE_CACHE is set, we use boost::uuids::detail::sha1 to compute the SHA1 of the program source, options, etc, and use it as the unique key to manage cached objects.

When the hash is converted to a string and passed to load_program_binary() and save_program_binary(), the conversion function of object detail::sha1 is called, which then calls  boost::uuids::detail::sha1::get_digest(). If you take a look at http://www.boost.org/doc/libs/1_63_0/boost/uuid/sha1.hpp, the sha1::get_digest() function actually appends some bytes to the string each time called. Thus, each time when we call get_digest() we get a different SHA1. 

Thus, the hash string used as the parameter to load_program_binary() and save_program_binary() are always different, because the string conversion function is called twice. This effectively disables the offline cache as the two hashes are always different.

My patch fixes this issue by converting the SHA1 to string only once, and use the same string for  load_program_binary() and save_program_binary().

### DIFF
--- a/include/boost/compute/program.hpp
+++ b/include/boost/compute/program.hpp
@@ -543,10 +543,11 @@ public:
             .process( options     )
             .process( source      )
             ;
+        std::string hash_string = hash;
 
         // Try to get cached program binaries:
         try {
-            boost::optional<program> prog = load_program_binary(hash, context);
+            boost::optional<program> prog = load_program_binary(hash_string, context);
 
             if (prog) {
                 prog->build(options);
@@ -575,7 +576,7 @@ public:
 
 #ifdef BOOST_COMPUTE_USE_OFFLINE_CACHE
         // Save program binaries for future reuse.
-        save_program_binary(hash, prog);
+        save_program_binary(hash_string, prog);
 #endif
 
         return prog;


haahh:
Good job! Thanks!

keryell:
Ouch!
What a nasty bug.
Perhaps this "feature" from https://github.com/boostorg/uuid should be fixed upstream instead?
Interestingly, issues are not allowed on this repository. :-)

Another fix could have been to change the logic of `boost/compute/detail/sha1.hpp` to add the caching of the returned `std::string`.

But I think it would be nice to add a few comments in the code explaining the issue and the change.
Otherwise it is impossible to understand why we put the hash string in an intermediate string and someone else could be bitten again in the future...

haahh:
>Perhaps this "feature" from https://github.com/boostorg/uuid should be fixed upstream instead?
Interestingly, issues are not allowed on this repository. :-)

<a href="https://imgflip.com/i/1mblq0"></a>

>Another fix could have been to change the logic of boost/compute/detail/sha1.hpp to add the caching of the returned std::string.

Not a bad idea.

```cpp
// Accumulates SHA1 hash of the passed strings.
class sha1 {
    public:
        sha1(const std::string &s = "") {
            if (!s.empty()) this->process(s);
            else build_hash_str = true;
        }

        sha1& process(const std::string &s) {
            build_hash_str = true;
            h.process_bytes(s.c_str(), s.size());
            return *this;
        }

        operator std::string() {
            if(!build_hash_str)
                return hash_str;

            unsigned int digest[5];
            h.get_digest(digest);

            std::ostringstream buf;
            for(int i = 0; i < 5; ++i)
                buf << std::hex << std::setfill('0') << std::setw(8) << digest[i];

            hash_str = buf.str();
            return hash_str;
        }
    private:
        boost::uuids::detail::sha1 h;
        std::string hash_str;
        bool build_hash_str;
};
```

but

> a few comments in the code explaining the issue and the change.
Otherwise it is impossible to understand why we put the hash string in an intermediate string and someone else could be bitten again in the future...

seems to be a simpler solution.

PR #701, opened 2017-04-02
radix sort by key with custom value type, fixes #162

### LINKED ISSUE #162: Problem using arbitrary struct as value parameter in compute::sort_by_key()
Hello Kyle,

I tried to use compute::sort_by_key with a vector<float> as key and a struct like that

```
typedef struct
{
    float a;
    float v[20];
} imx;
vector<imx>
```

as the value (or payload).
But unfortunately I cannot use this struct, only basic data types like int, float, double or so ...
Browsing the code it looks not too difficult to extend the current behaviour ...

/Jesko

```
//--------------------------------------------------------- code snippet below
    std::vector<float> host_vector_keys(33);
    std::vector<imx> host_vector_payload(host_vector_keys.size());
    compute::vector<float> device_vector_keys = host_vector_keys;
    compute::vector<imx> device_vector_payload = host_vector_payload;
    compute::sort_by_key(device_vector_keys.begin(), device_vector_keys.end(), device_vector_payload.begin());
```
##

haahh:
Based on https://github.com/boostorg/compute/pull/568 I prepared a fix for radix sort by key algorithm. Fixes #162.

In https://github.com/haahh/compute/commit/f1160bbc1f513224709b26a19875589260e63560 I did the same fix but using different approach. If you think it's better I'll make pull request with that one.

### DIFF
--- a/include/boost/compute/algorithm/detail/radix_sort.hpp
+++ b/include/boost/compute/algorithm/detail/radix_sort.hpp
@@ -307,7 +307,7 @@ inline void radix_sort_impl(const buffer_iterator<T> first,
 
     // load radix sort program
     program radix_sort_program = cache->get_or_build(
-        cache_key, options.str(), radix_sort_source, context
+       cache_key, options.str(), boost::compute::type_definition<T2>() +"\n"+radix_sort_source, context
     );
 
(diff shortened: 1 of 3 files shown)


kylelutz:
Looks good! Thanks!

PR #710, opened 2017-04-21
Add AppVayor CI

jszuppe:
This PR adds AppVayor CI and fixes some bugs in tests.

https://github.com/boostorg/compute/commit/2a6995581432ab602459aff81a1ecee6e175b732 is the most interesting bug-fix. `clEnqueueWriteBuffer` is not "100% blocking". OpenCL specification only promises that for `clEnqueueWriteBuffer`:

>if blocking_write is `CL_TRUE`, the OpenCL implementation copies the data referred to by `ptr` and enqueues the write operation in the command-queue. The memory pointed to by `ptr` can be reused by the application after the `clEnqueueWriteBuffer` call returns.

Which means that blocking `clEnqueueWriteBuffer` can return after it copies memory pointer by `ptr` to some temporary memory (on host), and it doesn't have to wait for the write to the buffer to be completed. (To be honest OpenCL specification is not 100% clear about that.) Anyway, in case of the Intel implementation for CPU, `clEnqueueWriteBuffer` does not wait for the data to be written into the buffer. Every other platform waits (AMD, NVIDIA, pocl), but the fact is that Intel implementation is in line with the specification. (See also https://software.intel.com/en-us/forums/opencl/topic/731519.)

That caused random errors when writing and reading using `[]` operator, because each call of `[]` operator creates it's own temporary command queue (so calls were not ordered). Long story short: It was possible to do `vector[0] = 3;` and right after that `vector[0] == 3` could be `false`.


haahh:
@kylelutz I guess we need some connection with AppVeyor CI in repository settings. I don't have permissions to do it. When it's activate, send me the code with the badge, so I'll update commit https://github.com/boostorg/compute/pull/710/commits/1e8d292478ca639e32aff6105d793a7e5b3f3037. It would be great if I can have access to the AppVeyor CI of Boost.Compute like I have in Travis CI.

Check https://ci.appveyor.com/project/haahh/compute-e97yn/build/1.0.1 to see how AppVeyor builds  work.

haahh:
I'll move fixes to separate pull request.

kylelutz:
Awesome! Merged!

PR #712, opened 2017-04-23
Optionally enforce device selected with BOOST_COMPUTE_DEFAULT variables

### LINKED ISSUE #692: Optionally enforce algorithm to pick the default device
When using some weird devices or debugging OpenCL stacks & compilers, it does not make sense to run the application on anything else.

If I correctly understand `find_default_device()`algorithm in https://github.com/boostorg/compute/blob/master/include/boost/compute/system.hpp#L210 on using the very handy`BOOST_COMPUTE_DEFAULT_` variables, when some criteria cannot be fulfilled we fall-back on returning the first GPU found, etc.

So I'd like to add another variable environment for example (`BOOST_COMPUTE_DEFAULT_ENFORCE`? Bike-sheding welcome! :-) ) to change the selection algorithm to throw when a `BOOST_COMPUTE_DEFAULT_` variable is set but the criteria cannot be fulfilled.

Any comment?

I can write a PR.

ddemidov:
Fixes #692

### DIFF
--- a/include/boost/compute/system.hpp
+++ b/include/boost/compute/system.hpp
@@ -55,6 +55,10 @@ public:
     ///        name of the platform (e.g. "NVIDIA CUDA")
     /// \li \c BOOST_COMPUTE_DEFAULT_VENDOR -
     ///        name of the device vendor (e.g. "NVIDIA")
+    /// \li \c BOOST_COMPUTE_DEFAULT_ENFORCE -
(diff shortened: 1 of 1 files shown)


haahh:
@keryell I think it's what you've wanted. I'll merge it later today.

haahh:
@ddemidov Thanks!

PR #717, opened 2017-05-07
Add get_sub_group_info() to kernel class

jszuppe:


jszuppe:
* Add `get_sub_group_info()` method
* Use latest OpenCL headers for AppVeyor CI
* Various other small fixes.

`get_sub_group_info()` method returns `boost::optional<T>`. That way, instead of throwing exception in user's face, we can just return null optional if device is not 2.1 device, or is not 2.0 device with support for `cl_khr_subgroups` extension.

jszuppe:
Still no support for AppVeyor in Github settings, so https://ci.appveyor.com/project/jszuppe/compute-e97yn.

PR #737, opened 2017-08-08
size_t and ptrdiff_t need special treatment on OSX

ddemidov:
The types do not resolve to `ulong` and `long` on OSX, as opposed to other platforms.
See ddemidov/vexcl#222

### DIFF
--- a/include/boost/compute/types.hpp
+++ b/include/boost/compute/types.hpp
@@ -20,5 +20,6 @@
 #include <boost/compute/types/pair.hpp>
 #include <boost/compute/types/struct.hpp>
 #include <boost/compute/types/tuple.hpp>
+#include <boost/compute/types/size_t.hpp>
 
 #endif // BOOST_COMPUTE_TYPES_HPP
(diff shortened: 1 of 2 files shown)


ddemidov:
Unfortunately, I don't have access to OSX, so I am not able to test this. @chrism0dwk, would you mind copying [compute/types/size_t.hpp](https://github.com/ddemidov/compute/blob/c9522ad4c0f231bbaa626164cd0741f5783a42d7/include/boost/compute/types/size_t.hpp) to your OSX machine and testing if the following  compiles:
```cpp
#include <iostream>
#include <boost/compute.hpp>
#include "size_t.hpp"
#define VEXCL_BACKEND_COMPUTE
#include <vexcl/vexcl.hpp>

int main() { }
```

chrism0dwk:
@ddemidov yes, your test program compiles on my OSX machine with the size_t.hpp file present on the include path.

ddemidov:
Travis fails, but it does not seem related to the PR.

PR #742, opened 2017-09-10
Fix strided_iterator tag

### LINKED ISSUE #741: std::distance on strided iterator runs forever
Minimal reproducible example on boost 1.64

```
boost::compute::vector<float> vec(300);
boost::compute::strided_iterator<boost::compute::buffer_iterator<float> > begin(vec.begin(),1);
boost::compute::strided_iterator<boost::compute::buffer_iterator<float> > end(vec.end(),1);
std::distance(end,begin);
```

This code will run forever. Note that this works perfectly on the buffer iterator alone and it works perfectly when switching begin and end. A rough debugging session reveals that the call above for some reason will choose the default version of std::distance which will iterate over the iterator range. For 64 bit this takes some time.

I came to this issue because for some reason for some iterators a call of begin.distanceTo(end) in the iterator base class is rewritten as -end.distanceTo(begin) somewhere deep in the call stack.

jszuppe:
Using `std::iterator_traits<>` to get `iterator_catagory` prevents from converting STL iterator tag to Boost tag.

Fixes #741

### DIFF
--- a/include/boost/compute/iterator/strided_iterator.hpp
+++ b/include/boost/compute/iterator/strided_iterator.hpp
@@ -42,7 +42,9 @@ class strided_iterator_base
 public:
     typedef ::boost::iterator_adaptor<
         ::boost::compute::strided_iterator<Iterator>,
-        Iterator
+        Iterator,
+        typename std::iterator_traits<Iterator>::value_type,
+        typename std::iterator_traits<Iterator>::iterator_category
     > type;
 };
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,7 +73,7 @@ matrix:
        - LINUX_DIST=trusty
        - OPENCL_LIB=pocl
        - OPENCL_HEADERS_VER="21"
-       - ENV_CXX_FLAGS="-DBOOST_COMPUTE_MAX_CL_VERSION=100"
+       - ENV_CXX_FLAGS="-Wno-unused-local-typedef -DBOOST_COMPUTE_MAX_CL_VERSION=100"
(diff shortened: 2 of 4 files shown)


include/boost/compute/iterator/strided_iterator.hpp
```diff
@@ -15,6 +15,7 @@
 #include <iterator>
 
 #include <boost/config.hpp>
+#include <boost/static_assert.hpp>
```
kylelutz: Doesn't look like this is used here, just included for the test?

kylelutz:
Looks good, left one comment. Thanks!

include/boost/compute/iterator/strided_iterator.hpp
jszuppe: Oh, sorry. I meant to include that in the test. I'll fix it in a moment.

jszuppe:
@kylelutz Do you have any idea why Boost.Compute's builds have such a bit delay on Travis CI? I have Travis CI activated for my fork and it always starts immediately: https://travis-ci.org/jszuppe/compute/builds/273862336

kylelutz:
@jszuppe Yeah, that's really strange. AFAICT, they both have the same `.travis.yml` and shouldn't really have any differences. Though I wonder if there's some org-wide throttling that queues us behind other `boostorg` builds?

jszuppe:
I updated the merge-request. You can check build on Travis CI here: https://travis-ci.org/jszuppe/compute/builds/273955520, just in case Boost.Compute build does not start soon.

> Though I wonder if there's some org-wide throttling that queues us behind other boostorg builds?

Maybe, but when I go into boostorg page it does not seem like there are a lot of builds. Also it looks like [this](https://travis-ci.org/boostorg/hana/builds/273421992) build in Hana lib was started just an 2 hours ago, and it has been started.

PR #747, opened 2017-10-12
Fix Threefry RNG

### LINKED ISSUE #745: threefry random number generator broken
Hi,
There are 2 issues.

1. Threefry does not adhere to the interface of generators. The following program will not compile:

```
#include <boost/compute/system.hpp>
#include <boost/compute/random/threefry_engine.hpp>
#include <boost/compute/random/normal_distribution.hpp>
#include <boost/compute/random/uniform_real_distribution.hpp>

namespace compute = boost::compute;
int main()
{
	// get default device and setup context
	compute::device gpu = compute::system::default_device();
	compute::context context(gpu);
	compute::command_queue queue(context, gpu);
	compute::threefry_engine<> random_engine(queue);
	
	compute::uniform_real_distribution<float> random_distribution(0.f, 4.f); //does not compile
	// compute::normal_distribution<float> random_distribution(0.f, 4.f); // compiles
	
	compute::vector<float> random_values(10, 0.0,queue);
	
	random_distribution.generate(
		random_values.begin(), random_values.end(), random_engine, queue
	);
}
```

2. Threefry does not have any source of entropy and adding the following to the code above will print out two times the same sequence:

```
    for(std::size_t i = 0; i != 2; ++i){
		compute::vector<compute::uint_> random_values(10, 0.0,queue);
		
		random_engine.generate(
			random_values.begin(), random_values.end(), queue
		);
		
		std::vector<compute::uint_> values_cpu(10);
		boost::compute::copy(random_values.begin(),random_values.end(), values_cpu.begin(), queue);
		
		for(std::size_t i = 0; i != 10; ++i){
			std::cout<<values_cpu[i]<<" ";
		}
		std::cout<<std::endl;
	}
```
3. This behaviour is not observed when generating normally distributed numbers from the generator. This is because the input sequence to the generator is not initialized and thus the generator relies on uninitialized memory, see https://github.com/boostorg/compute/blob/d36ef02b7860b9bcf5ea4911b77f1aa77b5f4059/include/boost/compute/random/normal_distribution.hpp#L91

How to fix this:
Threefry relies on its key as source of entropy and then generates the random numbers from a simple counting sequence 1,2,3.. . So an easy solution would be to use a mt19937 on cpu to generate a key and set it as kernel argument. This would also allow seeding through the mt19937. Then, within the kernel the function threefry2x32_R should be given the number of the current random number as first and the key as second argument. It is okay if the sequence starts every time from 0, as threefry would still be able to generate 2^64 different sequences. If this is not enough, either use threefry4x32_R for 2^128 different sequences or initialize the counter to some random offset.

jszuppe:
Fixes #745

### DIFF
--- a/include/boost/compute/random/threefry_engine.hpp
+++ b/include/boost/compute/random/threefry_engine.hpp
@@ -34,20 +34,27 @@ template<class T = uint_>
 class threefry_engine
 {
 public:
-    static const size_t threads = 1024;
     typedef T result_type;
+    static const ulong_ default_seed = 0UL;
 
     /// Creates a new threefry_engine and seeds it with \p value.
-    explicit threefry_engine(command_queue &queue)
-        : m_context(queue.get_context())
+    explicit threefry_engine(command_queue &queue,
+                             ulong_ value = default_seed)
+        : m_counter(0),
+          m_context(queue.get_context())
     {
-        // setup program
+        // Set seed
+        m_key[0] = static_cast<uint_>(value);
+        m_key[1] = static_cast<uint_>(value >> 32);
+        // Load program
         load_program();
     }
 
     /// Creates a new threefry_engine object as a copy of \p other.
     threefry_engine(const threefry_engine<T> &other)
-        : m_context(other.m_context),
+        : m_key(other.m_key),
+          m_counter(other.m_counter),
+          m_context(other.m_context),
           m_program(other.m_program)
     {
     }
@@ -56,6 +63,8 @@ public:
     threefry_engine<T>& operator=(const threefry_engine<T> &other)
     {
         if(this != &other){
+            m_key = other.m_key;
+            m_counter = other.m_counter;
             m_context = other.m_context;
             m_program = other.m_program;
         }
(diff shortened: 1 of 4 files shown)


jszuppe:
@Ulfgard Please review

jszuppe:
I'll focus on fixing Travis CI builds tomorrow.

include/boost/compute/random/threefry_engine.hpp
```diff
@@ -68,14 +77,81 @@ class threefry_engine
...
+    void generate(discard_iterator first, discard_iterator last, command_queue &queue)
+    {
+        (void) queue;
+        ulong_ offset = std::distance(first, last);
+
+        uint_ lo = static_cast<uint_>(offset);
+        uint_ hi = static_cast<uint_>(offset >> 32);
+
+        m_counter[0] += lo;
+        m_counter[1] += hi + (m_counter[0] < lo ? 1 : 0);
```
Ulfgard: consider making counter a 64bit integer just for making this logic simpler. This also allows to make counter an atomic for thread safety.

jszuppe: I can make it `ulong_` (and the `m_key` too), but I'm not sure about making the counter an atomic value. Other Boost.Compute RNGs do not promise thread safety (just like RNGs in C++ Standard Library).

Ulfgard:
lloks good. ship it :)

PR #748, opened 2017-10-16
Support for set_arg on macOS

henryiii:
Currently, macOS does not support set_arg with some datatypes. This was partially fixed in #737, but it does not work if the user does not manually specify the `<boost/compute/types.hpp>` header manually.

This PR is expected to fail to build on macOS on the first build. It will then need to be fixed. 

See ddemidov/vexcl#227 for more info.

### DIFF
--- a/include/boost/compute/core.hpp
+++ b/include/boost/compute/core.hpp
@@ -22,6 +22,7 @@
 #include <boost/compute/device.hpp>
 #include <boost/compute/event.hpp>
 #include <boost/compute/kernel.hpp>
+#include <boost/compute/types.hpp>
 #include <boost/compute/memory_object.hpp>
 #include <boost/compute/platform.hpp>
 #include <boost/compute/program.hpp>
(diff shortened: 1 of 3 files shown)


henryiii:
At least now you can see the macOS results without a Mac. :) (I do the same with Windows and AppVeyor)

henryiii:
Rebasing on master didn't seem to stop the build errors for Compute.

ddemidov:
@henryiii , can you try to add https://github.com/ddemidov/compute/commit/8f77efd0ef1aeeeed09fd097a675daa4eb9da644 to this PR? I believe that should fix the issue.

ddemidov:
I think the problem now is that [`test_kernels.cpp`](https://github.com/henryiii/compute/blob/70aa256ec439c0b72738e8e9cbfc8279f5c128c7/test/test_kernel.cpp#L14-L17) does not include `compute/core.hpp`. And we can not include `types.hpp` into the `kernel.hpp`, because `types.hpp` specializes some templates only declared in `kernel.hpp`. So may be the best solution for the failing test would be to either include `core.hpp` instead of `kernel.hpp`, or include `types.hpp` after `kernel.hpp`.

henryiii:
It bothers me that include order matters. Clang can even warn if includes are not sorted in alphabetical order.

Couldn't `types` include `kernel`, since it works on the contents?

henryiii:
I believe this fix is okay to add.

PR #765, opened 2018-02-20
add future::then() method

### LINKED ISSUE #94: Add future::then() method
Add a `future<T>::then()` method which takes a generic function callback and invokes it once the future is ready. Should be trivial once `event::set_callback()` supports generic function callbacks (see issue #86).
##

lakshayg:
resolves #94

### DIFF
--- a/include/boost/compute/async/future.hpp
+++ b/include/boost/compute/async/future.hpp
@@ -82,6 +82,21 @@ public:
         return m_event;
     }
 
+    #if defined(BOOST_COMPUTE_CL_VERSION_1_1) || defined(BOOST_COMPUTE_DOXYGEN_INVOKED)
+    /// Invokes a generic callback function once the future is ready.
+    ///
+    /// The function specified by callback must be invokable with zero arguments.
+    ///
+    /// \see_opencl_ref{clSetEventCallback}
+    /// \opencl_version_warning{1,1}
+    template<class Function>
+    future& then(Function callback)
+    {
+        m_event.set_callback(callback);
+        return *this;
+    }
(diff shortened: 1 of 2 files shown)


jszuppe:
Please rebase on `develop`.

jszuppe:
Would be great if you add tests for that.

include/boost/compute/async/future.hpp
```diff
@@ -82,6 +82,15 @@ class future
         return m_event;
     }
 
+    #if defined(BOOST_COMPUTE_CL_VERSION_1_1) || defined(BOOST_COMPUTE_DOXYGEN_INVOKED)
```
jszuppe: I think it would be useful to add comment that
>  The function specified by `callback` must be invokable with zero arguments (e.g. `callback()`).

lakshayg:
Sure, I can add tests. In which file should I add them?

jszuppe:
I think `test_event.cpp` is a good place for that.

jszuppe:
Shouldn't `then()` return reference to modified `future` (`*this`)? OpenCL doesn't guarantee order of execution for event callbacks, but, nonetheless, user may want to register multiple callback to a future. I'd also add `\see_opencl_ref{clSetEventCallback}` and `\opencl_version_warning{1,1}` to the description, so it looks like this:

```
/// Invokes a generic callback function once the future is ready.
//
/// The function specified by callback must be invokable with zero arguments.
///
/// \see_opencl_ref{clSetEventCallback}
/// \opencl_version_warning{1,1}
```

Otherwise, looks good.

lakshayg:
Done

PR #767, opened 2018-02-25
DOC: Fixed missing images

### LINKED ISSUE #621: Images are missing in documentation
Images about preformance are missing in boost page documentation http://www.boost.org/doc/libs/1_61_0/libs/compute/doc/html/boost_compute/performance.html

Kojoley:
Images fetched from `gh-pages` branch.

Closes #621

Update: If you do not want to place images in `html` folder it can be lifted to `doc` with adding `<format>html:<xsl:param>img.src.path=../` to jamfile. But I do not know how it would affect your GH pages.


Kojoley:
I am closing this as there is no response

keryell:
Was it fixed?

Kojoley:
No, but everyone here seems to be fine with that

PR #772, opened 2018-04-10
Fix out of bound access in local_scan kernel

umar456:
The `local_scan` kernel was accessing the first buffer using the global id without checking if it was within the bounds.


umar456:
I found this by running cuda-memcheck. Is there anything else I should do to get this merged?

PR #773, opened 2018-04-11
Use system-dependent CL<->GL sharing extension name instead of hardcoded string

sznaider:
It looks as though the author's intention was to use different extension name for different platforms. 

For some reason, however, commit 114e444a7ea88055460f8dd76c6cc76032df6b18 introduced a change that ignores the platform-specific `cl_gl_sharing_extension` variable and uses a string instead, whose value doesn't seem right (and it doesn't work on my system).


jszuppe:
Thanks!

PR #776, opened 2018-05-08
Fix compilation errors with GCC 8

9prady9:

PR #782, opened 2018-05-18
const correct get_context() in svm_ptr

richard-vock:
gcc 8.1 now considers this an error.


jszuppe:
Thanks, but it's been already solved in https://github.com/boostorg/compute/pull/776. Also, pull request should be against `develop` branch.

PR #794, opened 2018-08-20
Add complex support to plus operator

bwitherspoon:


jszuppe:
I thought it's already implemented since it works with reduce: https://github.com/boostorg/compute/blob/master/test/test_reduce.cpp#L259

bwitherspoon:
Oh I see. I think I was mistakenly testing it against my system installed Boost 1.66 instead of master where it does seem to have already been implemented. Closing.

bwitherspoon:
Not sure what commit between the 1.66 tag and master permits that test to compile though.

PR #797, opened 2018-10-17
Added wait_list parameter to "copy" functions

u-s:
It's not possible to use current "copy" functions with CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE option in command queue, because wait_list parameter is missing.

### DIFF
--- a/include/boost/compute/algorithm/copy.hpp
+++ b/include/boost/compute/algorithm/copy.hpp
@@ -103,6 +103,7 @@ dispatch_copy_async(InputIterator first,
                     InputIterator last,
                     OutputIterator result,
                     command_queue &queue,
+                    const wait_list &events,
                     typename boost::enable_if<
                         mpl::and_<
                             mpl::not_<
@@ -118,7 +119,7 @@ dispatch_copy_async(InputIterator first,
         "copy_async() is only supported for contiguous host iterators"
     );
 
-    return copy_to_device_async(first, last, result, queue);
+    return copy_to_device_async(first, last, result, queue, events);
 }
 
 // host -> device (async)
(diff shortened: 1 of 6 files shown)


keryell:
Quite pervasive but it looks good,

jszuppe:
Wouldn't it be better if it was done only for `copy_async`? Because, honestly, what's the help if you use `copy` which does not give you an event with out-of-order queue?

Ulfgard:
It still helps. if you have an OoO-Queue, you might still want a blocking write operation ("copy back to device after you are done computing") but this has to wait until all relevant operations are done. This is especially interesting in a multi-thread context where each thread enqueues items independently into the same queue to ensure that the device always has enough to work on. In this case, doing explicit synchronization of the queue is harmfull to performance, synchronous copy is not.
________________________________

jszuppe:
Right, I just don't want to run into a situation where we have multiple different ways async API is done in Boost.Compute. That's why I am just wondering if it's maybe better to have only `*_async` functions to work with OoO queues for now. However, I'm not saying 'no' to this change. @kylelutz What do you think?

Anyway, I would need to revive our CI and test it on a few platforms (I think full coverage is: Mac, NV, AMD, Intel, and POCL; IMHO can be limited to 1.2) to make sure it's all right. If anyone can help with that I'll be grateful.

Ulfgard:
Without a wait_list, synchronous operations are impossible to use safely in an OoO context. In OpenCL, all operations except copies are asynchronous and synchronous copies have to wait until those operations are done. This little detail is abstracted away in an in-order queue since there is an implicit ordering of events(i.e. wait_list=all currently unfinished operations). In OoO we have to make this ordering explicit, otherwise the copy takes place immediately once the memory bus is free.

Issue #774, opened 2018-04-16
Wrong sort results using customized compare


antibes0415:
Hi. I want to sort 64-bit long variates according to their upper 32 bit value.
Here is the code:

`  
BOOST_COMPUTE_FUNCTION(bool, compareSampleKeys, (cl_ulong a, cl_ulong b), {
    return (a >> 32) < (b >> 32);
  });
  // cl::Buffer *m_dev_sampleKeys;
  boost::compute::buffer cbSampleKeys((*m_dev_sampleKeys)());

  boost::compute::buffer_iterator<cl_ulong> iter =
    boost::compute::make_buffer_iterator<cl_ulong>(cbSampleKeys, 0);

  boost::compute::sort(
    iter,
    iter + m_nbSamples,
    compareSampleKeys,
    m_compute_queue
  );
`

Don't know why lots values are modified after sorting. In my case, all numbers are unique and there turn out to be many repetitive numbers after sorting. I tried not using customized compare, and that won't modify the values.

rosenrodt:
I have the same issue with compute::sort() too. I tried a vector of custom struct. Unique keys in, duplicate keys out.

jszuppe:
Fixed in #803

PR #803, opened 2018-12-19
Fix duplicates using custom compare with bitonic block sort

rosenrodt:
I have modified test case to reproduce the issue as described in #774. And a fix is proposed

Tested on:
* Nvidia Geforce GTX1060 
* AMD Radeon HD 6770

### DIFF
--- a/include/boost/compute/algorithm/detail/merge_sort_on_gpu.hpp
+++ b/include/boost/compute/algorithm/detail/merge_sort_on_gpu.hpp
@@ -170,8 +170,12 @@ inline size_t bitonic_block_sort(KeyIterator keys_first,
             k.decl<bool>("compare") << " = " <<
                 compare(k.var<key_type>("sibling_key"),
                         k.var<key_type>("my_key")) << ";\n" <<
+            k.decl<bool>("equal") << " = !(compare || " <<
+                compare(k.var<key_type>("my_key"),
+                        k.var<key_type>("sibling_key")) << ");\n" <<
             k.decl<bool>("swap") <<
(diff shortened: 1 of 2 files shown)


jszuppe:
Thanks! I'll look into today or later this week and merge.

jszuppe:
It would be great if you can change the order of commits: first test which shows the bug, then commit with a fix. Please reference issue id in the fix commit. If you have problems with doing that, or don't know how to do it using git, don't worry, I'll do that for you. Thanks again for submitting a fix and testing it on two platforms!

rosenrodt:
@jszuppe thanks. i've rearranged the commits

jszuppe:
After looking at it.

1. Will it work correctly if we do opposite: force them not to do the swap if they are equal? This is always one local memory operation less. I think the bug is/was that if there are equal values one is swapped and one is not.
2. If 1. does not work, then I think we can at least limit your fix to `sort_by_key` situation only.

rosenrodt:
1. That will do too. I have pushed them too in the PR 
2. The fix concerns `bitonic_block_sort` with `sort_by_key = false`. So I am not sure what you mean by that. Although I haven't looked into it but does the CPU equivalent or `bitonic_block_sort` with `sort_by_key = true` could possibly repro issue #774 too?

btw the Travis build server seems to complain about missing `CL.h`

jszuppe:
1. Ok
2. Yeah, you're right. I think it can apply sort-by-key situation too with keys, but of course this fixes also that problem.

jszuppe:
> btw the Travis build server seems to complain about missing CL.h

Yeah, I'll have to fix Travis CI script someday.... Too much work at work ;)

Issue #811, opened 2019-01-10
Sort algorithms fail running on AMD Radeon RX Vega 56


rosenrodt:
Similar to issue reported in #795, I am facing all kinds of sort() failures with AMD RX Vega GPUs

I ran the test with the command `ctest --output-on-failure`. And here is the summary 

### Test result for driver Adrenalin 18.5.2 & 18.9.1
The following tests FAILED:
         54 - algorithm.radix_sort (Failed)
         55 - algorithm.radix_sort_by_key (Failed)
         75 - algorithm.sort_by_key (Failed)
         77 - algorithm.stable_sort (Failed)
        143 - misc.amd_cpp_kernel_language (Failed)
        148 - example.amd_cpp_kernel (Exit code 0xc0000409)

### Test result for driver Adrenalin 18.12.2   
The following tests FAILED:
         **41 - algorithm.insertion_sort (Failed)**
         **45 - algorithm.merge_sort_gpu (Failed)**
         54 - algorithm.radix_sort (Failed)
         55 - algorithm.radix_sort_by_key (Failed)
         **74 - algorithm.sort (Failed)**
         75 - algorithm.sort_by_key (Failed)
         77 - algorithm.stable_sort (Failed)
        143 - misc.amd_cpp_kernel_language (Failed)
        148 - example.amd_cpp_kernel (Exit code 0xc0000409)

**Bold** items mean the test failed on new driver but not in older drivers

Curiously, with latest drivers it gets even worse. 

Note 1. For complete failure reports look [here](https://gist.github.com/rosenrodt/c801be1973d4b41a700a2cd100bc5ef2).   
Note 2. As I recall, AMD Radeon HD 6770 passes every test (maybe except for the amd_cpp_kernel tests)

rosenrodt:
Back with some quick update: Adrenalin 18.12.2 doesn't seem like a quality driver so let's just ignore it. I believe this issue is caused by pointer aliasing inside radix sort kernels. Will post a PR as soon as it is been confirmed

jszuppe:
Which pointers are aliasing each other?

rosenrodt:
I figured it’s not pointer aliasing in the scan() kernel I checked (see the work-in-progress pr #812). But rather variable dependency is not detected by the AMD OpenCL compiler. 

[...]

PR #812, opened 2019-01-15
fixes radix sort error

rosenrodt:
insert memory fence to avoid AMD OpenCL compiler reordering the read/writes; fixes issue #811

CTest log for Vega56

```
99% tests passed, 2 tests failed out of 169
 
Total Test time (real) =  85.07 sec

The following tests FAILED:
 	143 - misc.amd_cpp_kernel_language (Failed)
 	148 - example.amd_cpp_kernel (Exit code 0xc0000409)
Errors while running CTest
```

### DIFF
--- a/include/boost/compute/algorithm/detail/radix_sort.hpp
+++ b/include/boost/compute/algorithm/detail/radix_sort.hpp
@@ -176,8 +176,10 @@ const char radix_sort_source[] =
 "    uint sum = 0;\n"
 "    for(uint i = 0; i < K2_BITS; i++){\n"
 "        uint x = global_offsets[i] + last_block_offsets[i];\n"
+"        mem_fence(CLK_GLOBAL_MEM_FENCE);\n" // work around the RX 500/Vega bug, see #811
 "        global_offsets[i] = sum;\n"
 "        sum += x;\n"
+"        mem_fence(CLK_GLOBAL_MEM_FENCE);\n" // work around the RX Vega bug, see #811
 "    }\n"
 "}\n"
 


rosenrodt:
I am closing this PR because some issues still spill over after a clean rebuild on my test machine. Will reopen when it's done

rosenrodt:
I closed and reopened the PR because I was unsure of the cause of failed assertions when invoking `is_sorted()` on sorted arrays, but later figured it was totally unrelated to radix sort. Details were described in my comment [here](https://github.com/boostorg/compute/issues/811#issuecomment-454347805)

The PR is to make sure the memory access is in correct order and not mangled up by the compiler, therefore the `mem_fence` in [line 179](https://github.com/boostorg/compute/pull/812/files#diff-a3cb7a47c4aebcbfb2d5c01dca8995f6R179). But I rarely encounter such a case where compiler is not respecting obvious variable dependencies. So this is just an ad hoc fix until maybe we can get AMD to look into this, especially by the fact that it only happens on AMD Vega/RX5xx and not the older chips nor chips from other vendors. 

Another `mem_fence` is inserted for no apparent reason because it has to be like this to work on Vegas.

jszuppe:
Hmm.. At least a fence for global memory would make more sense, does that option work too?

rosenrodt:
Fences for global, local or both also work. So it won't make a difference. But I think you're right on that: global fencing is more appropriate in literal sense

jszuppe:
If I'm correct this kernel is run with one thread (via `enqueue_task`), so those fences should not have any performance cost. Because of that it shouldn't be a problem to include that workaround after:
* You change fence flag to global memory
* In those lines add reference to #811 bug, for example `// workaround for Radeon RX Vega bug, see #811`.

rosenrodt:
bump 😃 
@jszuppe anything else that needs to be done?

jszuppe:
No, it's ok. I just didn't have time to look at it.

PR #817, opened 2019-01-22
Work around incomplete closure input arguments on some platforms

rosenrodt:
This fixes all the kernel compile errors I encountered when running `test_closure`, one example:

```
--- build log ---
BC-src-code:6:75: error: use of undeclared identifier 'pi'
 inline float add_two_and_pi(float x, int two, float  p){ return x + two + pi; }
                                                                           ^
1 diagnostic(s) generated.
```

I encountered this curious case when cross-compiling with the standalone toolchain generated from [android NDK](https://developer.android.com/ndk/guides/standalone_toolchain). Most tests pass. But closure is among one of the few that fails. It seems like boost::trim() from the main boost library somehow cuts off more character than it should and also leaves a leading whitespace. 

```log
two, pi          // original string   
[two,  pi]       // vector after split()

two              // after trim()
int two          // capture_traits<T>::type_name() << " " << variable_name

 p               // after trim(), there is a leading space before 'pi' and 'i' is cut off 
float  p         // capture_traits<T>::type_name() << " " << variable_name
```

### DIFF
--- a/include/boost/compute/closure.hpp
+++ b/include/boost/compute/closure.hpp
@@ -216,7 +216,7 @@ struct closure_signature_argument_inserter
         BOOST_ASSERT(capture_string[0] == '(' &&
                      capture_string[capture_string_length-1] == ')');
         std::string capture_string_(capture_string + 1, capture_string_length - 2);
-        boost::split(m_capture_names, capture_string_ , boost::is_any_of(","));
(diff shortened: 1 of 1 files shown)


jszuppe:
That's odd that such an old function works incorrectly (`boost::trim()`). You should report that to https://github.com/boostorg/algorithm.

Issue #828, opened 2019-04-02
Support setting kernel argument as NULL


rosenrodt:
As per Khronos OpenCL spec:
> If the argument is declared to be a pointer of a built-in scalar or vector type, or a user defined structure type in the global or constant address space, the memory object specified as argument value must be a buffer object **(or NULL)**.

I tried to set kernel argument like this 
```c++
some_kernel.set_arg(0, NULL);
```
but get error `invalid argument size` instead.

Because OpenCL explicitly allows NULL as input, I went on to see why and now maybe I've identified the problem:

```c++
// set_kernel_arg implementation for built-in types (including NULL)
template<class T>
struct set_kernel_arg
{
    typename boost::enable_if<is_fundamental<T> >::type
    operator()(kernel &kernel_, size_t index, const T &value)
    {
        kernel_.set_arg(index, sizeof(T), &value);
//                                        ^^^^^^
//                     now we are getting the address of NULL 
//                     certainly CL API won't allow an arbitrary address
    }
};
```
So Boost Compute needs somehow to identify the NULL, maybe via a special type inside `boost::compute`, and set the value NULL directly to the CL API instead of its address.

jszuppe:
I don't think it's a valid option. I can undestand why someone would like to pass a null pointer when the argument isn't requied or won't be used in OpenCL C API, but in Boost.Compute's API you would have to pass an uninitialized buffer object.

Also I don't see good solution to support that (using `NULL` is ` void set_arg(size_t index, const T &value)`, because `NULL` does not have its own type. 

User can always use `void set_arg(size_t index, size_t size, const void *value)` which will work correctly with `NULL`. Another way would be to specify the type

rosenrodt:
> User can always use void set_arg(size_t index, size_t size, const void *value) which will work correctly with NULL.

I wasn't aware of that. That should solve my issue, but felt using something like `std::nullptr_t` or `compute::null_t` would be neater. This way it also plays nice with the variadic `set_args(T&&... args)`

jszuppe:
We can add overload for `std::nullptr_t`, which will fix it for C++11 and newer, so it should have in variadic version.

PR #829, opened 2019-04-03
add support to set kernel argument as NULL; fixes issue #828

rosenrodt:
Per Khronos OpenCL spec:
>If the argument is declared to be a pointer of a built-in scalar or vector type, or a user defined structure type in the global or constant address space, the memory object specified as argument value must be a buffer object (or NULL).

But the following call causes error:
```c++
some_kernel.set_arg(0, NULL);
```
One way as suggested in #828 is to use
```c++
some_kernel.set_arg(0, sizeof(cl_mem), NULL);
```
Now it's just
```c++
some_kernel.set_arg(0, nullptr);
```

### DIFF
--- a/include/boost/compute/kernel.hpp
+++ b/include/boost/compute/kernel.hpp
@@ -363,6 +363,11 @@ public:
     /// // set argument to a local buffer with storage for 32 float's
     /// kernel.set_arg(0, local_buffer<float>(32));
     /// \endcode
+    ///
+    /// For setting NULL to global and constant memory arguments (C++11):
+    /// \code
+    /// kernel.set_arg(0, nullptr);
+    /// \endcode
     template<class T>
     void set_arg(size_t index, const T &value)
     {
@@ -370,6 +375,14 @@ public:
         // attempted to set a kernel argument from an invalid type.
         detail::set_kernel_arg<T>()(*this, index, value);
     }
+
+    #ifdef BOOST_COMPUTE_USE_CPP11
+    /// \overload
+    void set_arg(size_t index, std::nullptr_t nul)
+    {
+        set_arg(index, sizeof(cl_mem), NULL);
+    }
+    #endif
 
     /// \internal_
     void set_arg(size_t index, const cl_mem mem)
--- a/include/boost/compute/kernel.hpp
+++ b/include/boost/compute/kernel.hpp
@@ -376,13 +376,13 @@ public:
         detail::set_kernel_arg<T>()(*this, index, value);
     }
 
-    #ifdef BOOST_COMPUTE_USE_CPP11
+    #ifndef BOOST_NO_CXX11_NULLPTR
     /// \overload
     void set_arg(size_t index, std::nullptr_t nul)
     {
         set_arg(index, sizeof(cl_mem), NULL);
     }
-    #endif
+    #endif // BOOST_NO_CXX11_NULLPTR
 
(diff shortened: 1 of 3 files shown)


rosenrodt:
Maybe it's better if this supports C++11. Should we implement similar class to `local_buffer` (say `null_buffer`) to support setting kernel argument to NULL?

jszuppe:
> Maybe it's better if this supports pre-C++11. Should we implement similar class to local_buffer (say null_buffer) to support setting kernel argument to NULL?

I would rather have just the C++11 version. It's very unlikely that someone using C++03 will find and use `boost::compute::null_buffer` before just using `set_arg(0, sizeof(cl_mem), NULL);` (not mentioning that setting a buffer to `NULL` is not a popular use case).

rosenrodt:
We could stick to C++11. The trade-off would be that people using pre-C++11 won't be able to write nice things like `kernel.set_args(b0, b1, nullptr)`.

If that's indeed where we want to go then I'll update the PR.

jszuppe:
pre-C++11 you don't have variadic templates, you can't write `kernel.set_args(b0, b1, nullptr)` in C++03.

rosenrodt:
I'm sold. I'll just stick with C++11 version. Please review

include/boost/compute/kernel.hpp
```diff
@@ -371,6 +376,14 @@ class kernel
         detail::set_kernel_arg<T>()(*this, index, value);
     }
 
+    #ifdef BOOST_COMPUTE_USE_CPP11
```
jszuppe: We should be able to replace that with `#ifndef BOOST_NO_CXX11_NULLPTR`.

include/boost/compute/kernel.hpp
```diff
@@ -371,6 +376,14 @@ class kernel
...
+    /// \overload
+    void set_arg(size_t index, std::nullptr_t nul)
+    {
+        set_arg(index, sizeof(cl_mem), NULL);
+    }
+    #endif
```
jszuppe: `#endoif //  BOOST_NO_CXX11_NULLPTR`

test/test_kernel.cpp
```diff
@@ -70,6 +70,11 @@ BOOST_AUTO_TEST_CASE(set_buffer_arg)
 
     foo.set_arg(0, x);
     foo.set_arg(1, y.get());
+#ifdef BOOST_COMPUTE_USE_CPP11
```
jszuppe: `BOOST_NO_CXX11_NULLPTR`

include/boost/compute/memory/local_buffer.hpp
```diff
@@ -88,4 +88,4 @@ struct set_kernel_arg<local_buffer<T> >
 } // end compute namespace
 } // end boost namespace
 
-#endif // BOOST_COMPUTE_MEMORY_SVM_PTR_HPP
+#endif // BOOST_COMPUTE_MEMORY_LOCAL_BUFFER_HPP
```
jszuppe: Good catch.

include/boost/compute/kernel.hpp
rosenrodt: added in new commit

include/boost/compute/kernel.hpp
rosenrodt: added in new commit

Issue #827, opened 2019-03-31
Support user-defined default context/device


rosenrodt:
Sometimes it is desirable to have CL-accelerated libraries initialize with user context managed by some outside application. But Boost.Compute currently hides global state of `context/device/queue` from users; there is no way to let user override e.g., their own global default context.

Because Boost.Compute already maintains global state of `context/device/queue` inside static functions, I think the best case scenario is being able to assign global CL context (derived from preferred CL compute device) without relying on parsing through pre-defined environment variables which `default_context()` implements.

One of the main benefits is to make direct access of device memory between shared objects/DLLs easily achievable. When some main application initialize two libraries `Foo` and `Bar`, those two libraries also initialize all internal buffers with global user-defined context. That way, data produced by library `Foo` can be directly consumed by library `Bar`. If not being able to set global state it would take much effort for library authors to add additional input argument for virtually every Boost.Compute call.

rosenrodt:
I have a [WIP patch](https://github.com/boostorg/compute/compare/master...rosenrodt:pr-set-default-context) that lets you set global context at the very first call to `system::default_context()`, via an additional input argument `user_context`. Afterwards the default queue will be derived from the user device (which is in turn derived from user context) and the provided user context itself. It does not change the old behavior if additional input argument is not provided. 

This gives the same guarantee that the global states is only set when `default_device/context/command_queue()` is called for the first time. Any subsequent call afterwards just returns the same value over again. 

It does not yet guarantee thread safety for the first call. They have to be done atomically or it will have UB when threads concurrently reading and writing the static variables at the same location. Maybe I could add a lock guard inside the calls but it may take a performance hit.

jszuppe:
The idea overall is ok. However, I think having something like `set_default_context()` or `switch_default_context()` would be better, because you can return the old default context. Implementation may be less straighforward compared to your patch. 

>It does not yet guarantee thread safety for the first call. They have to be done atomically or it will have UB when threads concurrently reading and writing the static variables at the same location. Maybe I could add a lock guard inside the calls but it may take a performance hit.

Right, I think that would be a good idea. We don't want to break someone.

rosenrodt:
> The idea overall is ok. However, I think having something like set_default_context() or switch_default_context() would be better, because you can return the old default context. Implementation may be less straighforward compared to your patch.

[...]

PR #831, opened 2019-04-06
Allow attaching user's own OpenCL context to Boost.Compute

rosenrodt:
This PR allows user to attach their own OpenCL context via `default_context(context* = 0)`, with such benefits as described in #827. It adheres to `default_context()` behavior defined in boost/compute/system.hpp :

> The default context is created once on the first time this function is called. Calling this function multiple times will always result in the same context object being returned.

Now user can either let Boost.Compute decide which context to initialize with:
```c++
MyLib::MyLib(context ctx)
{
    boost::compute::default_context(); // optional
}
```
Or attach their own context:
```c++
MyLib::MyLib(context ctx)
{
    boost::compute::default_context(&ctx);
}
```

Test result of repeatedly attaching user context concurrently will correctly return only one unique context. See `test_user_context_thread_safety.cpp`:
```
$ ctest -C Release -R \w*user_default_context_thread_safety --repeat-until-fail 512

1/1 Test #16: 
Start 16: core.user_default_context_thread_safety
core.user_default_context_thread_safety ...   Passed    0.06 sec
# … repeat 512 times

100% tests passed, 0 tests failed out of 1

Total Test time (real) =  31.65 sec
```
Tests passed with the following flags:
BOOST_COMPUTE_THREAD_SAFE=ON, BOOST_COMPUTE_USE_CPP11=ON
BOOST_COMPUTE_THREAD_SAFE=ON, BOOST_COMPUTE_USE_CPP11=OFF


rosenrodt:
I am closing this while preparing to open another PR

PR #832, opened 2019-04-21
Allow manually setting default command queue via `default_queue()`

rosenrodt:
By doing so, the associated default device and context will automatically match the user-provided command queue. It

Guarantees unique default queue/device/context is returned when calling from multiple threads (requires config macro BOOST_COMPUTE_THREAD_SAFE)

Resolves issue #827

### DIFF
--- a/include/boost/compute/system.hpp
+++ b/include/boost/compute/system.hpp
@@ -17,6 +17,18 @@
 
 #include <boost/throw_exception.hpp>
 
+#if defined(BOOST_COMPUTE_THREAD_SAFE) 
+#  if defined(BOOST_COMPUTE_USE_CPP11)
+#    include <mutex>
+#    include <thread>
+#    include <atomic>
+#  else
+#    include <boost/thread/mutex.hpp>
+#    include <boost/thread/lock_guard.hpp>
+#    include <boost/atomic.hpp>
+#  endif
+#endif 
+
 #include <boost/compute/cl.hpp>
 #include <boost/compute/device.hpp>
 #include <boost/compute/context.hpp>
@@ -24,6 +36,7 @@
 #include <boost/compute/command_queue.hpp>
 #include <boost/compute/detail/getenv.hpp>
 #include <boost/compute/exception/no_device_found.hpp>
+#include <boost/compute/exception/context_error.hpp>
 
 namespace boost {
 namespace compute {
@@ -78,9 +91,7 @@ public:
     /// \endcode
     static device default_device()
     {
-        static device default_device = find_default_device();
-
-        return default_device;
+        return init_default_device();
     }
 
     /// Returns the device with \p name.
@@ -149,17 +160,21 @@ public:
     /// same context object being returned.
     static context default_context()
     {
-        static context default_context(default_device());
-
-        return default_context;
+        return init_default_context();
     }
 
     /// Returns the default command queue for the system.
-    static command_queue& default_queue()
+    ///
+    /// If user-provided command queue is given, the system-wide default context
+    /// and default device will be set up appropriately so that the default queue
+    /// matches the default context and device.
+    ///
+    /// The default queue is created once on the first time this function is
+    /// called. Calling this function multiple times will always result in the
+    /// same command queue object being returned.
+    static command_queue& default_queue(command_queue* user_queue = 0)
     {
-        static command_queue queue(default_context(), default_device());
-
-        return queue;
+        return init_default_queue(user_queue);
     }
 
     /// Blocks until all outstanding computations on the default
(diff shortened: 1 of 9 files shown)


rosenrodt:
There seems to have missing pthread linkage when compiling with `std::thread` on some machines (reported by travis-ci). But on my Linux virtual machine I couldn't repro the same issue. So I think maybe if updating to newer compiler helps?

rosenrodt:
@jszuppe any comment on it? 😃

include/boost/compute/system.hpp
```diff
@@ -280,6 +295,145 @@ class system
...
+    static void init_default_device_and_context(command_queue* user_queue)
+    {
+        device user_device = user_queue->get_device();
+        context user_context = user_queue->get_context();
+
+        if ( (user_device != init_default_device(&user_device)) ||
+             (user_context != init_default_context(&user_context)) )
+        {
+            // Try invoking default_queue() before anything else
+            BOOST_THROW_EXCEPTION(context_error(&user_context,
```
jszuppe: I'm not sure if `context_error` is right error here.  I think it should be custom exception similar in design to `no_device_found`.

include/boost/compute/system.hpp
```diff
@@ -149,17 +160,21 @@ class system
...
     /// Returns the default command queue for the system.
-    static command_queue& default_queue()
+    ///
+    /// If user-provided command queue is given, the system-wide default context
+    /// and default device will be set up appropriately so that the default queue
+    /// matches the default context and device.
+    ///
+    /// The default queue is created once on the first time this function is
+    /// called. Calling this function multiple times will always result in the
+    /// same command queue object being returned.
```
jszuppe: I would add information about potential exception if default context and.or default device don't match with device and context of `user_queue`.

jszuppe: You can point to the test you created.

test/test_attach_user_queue_error.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    compute::device user_device = compute::system::devices().front();
+    compute::context user_context(user_device);
+    compute::command_queue user_queue(user_context, user_device);
+
+    // Don't call default_device() or default_context() before calling 
+    // default_queue(command_queue* = 0) if you wish to attach your command queue
+    compute::system::default_device(); 
+
+    try
```
jszuppe: Use `BOOST_CHECK_THROW`

test/test_system.cpp
```diff
@@ -14,25 +14,48 @@
...
+    compute::device device = compute::system::default_device();
+    compute::context context = compute::context(device);
+
+    compute::command_queue queue1(context, device);
+    {
+        compute::system::default_queue(&queue1);
+        compute::command_queue default_queue = compute::system::default_queue();
+        BOOST_ASSERT(queue1 == default_queue);
+    }
+#ifdef NDEBUG
```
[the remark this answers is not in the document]
jszuppe: Aaa, ok, I guess we don't want to have asserts.

include/boost/compute/system.hpp
```diff
@@ -149,17 +160,21 @@ class system
...
+    static command_queue& default_queue(command_queue* user_queue = 0)
```
jszuppe: Shouldn't `user_queue` be `const`? Also have we/you considered using constant references instead of pointers?

jszuppe: This also applies to other functions which take pointer to queue/device/context.

jszuppe:
I'm really sorry for delay. I had a busy week and then a few days of holidays without PC.

include/boost/compute/system.hpp
rosenrodt: Glad you pointed that out. The original design was to set user-provided context so it sort of fits in. But now it's been user-provided command queue already. I think we can throw a more explicit exception `set_default_queue_error`

test/test_system.cpp
```diff
@@ -14,25 +14,48 @@
...
+BOOST_AUTO_TEST_CASE(user_default_queue)
+{
+    compute::device device = compute::system::default_device();
+    compute::context context = compute::context(device);
+
+    compute::command_queue queue1(context, device);
+    {
+        compute::system::default_queue(&queue1);
+        compute::command_queue default_queue = compute::system::default_queue();
+        BOOST_ASSERT(queue1 == default_queue);
```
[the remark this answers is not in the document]
rosenrodt: Agreed. Will revise soon this week

test/test_system.cpp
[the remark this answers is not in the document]
rosenrodt: Yep :)

include/boost/compute/system.hpp
rosenrodt: Agreed. Will revise soon this week

test/test_attach_user_queue_error.cpp
rosenrodt: Agreed

include/boost/compute/system.hpp
rosenrodt: I agree it should be made `const`. As for constant reference, I don't think it's possible because the intent is to double `default_queue()` as both means to get/set the default queue. And we don't have things like null reference. 

That leaves us with two options:
1. We spin out the "set" function which takes const reference
2. Keep `default_queue()` and make the input pass-by-value, with default value set to be a null command queue

rosenrodt:
@jszuppe Glad to hear you back from your holiday. I have left some reply under your comments and I am eager to start working on the feature again

include/boost/compute/system.hpp
jszuppe: https://github.com/boostorg/compute/blob/master/include/boost/compute/command_queue.hpp#L108 This ctor creates "null" command queue.

include/boost/compute/system.hpp
jszuppe: The name of the error sounds ok.

test/test_system.cpp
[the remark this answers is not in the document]
rosenrodt: Yeah that should work. Didn't figure that out myself

include/boost/compute/system.hpp
rosenrodt: Yeah that should work. Didn't figure that out myself

rosenrodt:
I just force-pushed a one line fix for test_attach_user_queue_error.cpp due to CI check failure

```c++
compute::device user_device = compute::system::devices().front();
compute::context user_context(user_device);
compute::command_queue user_queue(user_context, user_device);

//compute::system::default_device();
compute::system::default_context(); 
// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// purposefully set the default context that is incompatible with user-provided command queue

BOOST_CHECK_THROW(
    compute::system::default_queue(user_queue),
    compute::set_default_queue_error);
```

jszuppe:
Fails in CI are unrelated to the PR. Merging. Good job!

PR #834, opened 2019-05-09
Allow kernel argument to take custom struct by value

rosenrodt:
By specializing `template<T> struct set_kernel_arg` when expanding BOOST_COMPUTE_ADAPT_STRUCT, kernel input argument can now take custom struct by value. This potentially utilizes constant memory of the compute device


rosenrodt:
It looks like [Apple platform](https://travis-ci.org/boostorg/compute/jobs/530315064) regards custom struct as invalid kernel argument. That's weird because my 2016 Macbook works just fine

PR #867, opened 2021-08-19
lru_cache, remove redundant copy when updating queue position

Hailios:


include/boost/compute/detail/lru_cache.hpp
```diff
@@ -97,9 +97,7 @@ class lru_cache
             m_list.push_front(key);
 
             // update iterator in map
-            j = m_list.begin();
-            const value_type &value = i->second.first;
-            m_map[key] = std::make_pair(value, j);
+            i->second.second = m_list.begin();
 
             // return the value
             return value;
```
faker2021, line 103: Where is the value from.

Hailios: thanks, I forgot to update it. updated the PR now :)

PR #876, opened 2022-08-03
Create SECURITY.md

PixelMedia1:


striezel:
@PixelMedia1:
The file that is added in this pull requests looks like a placeholder and not like a real `SECURITY.md` file. A real file should actually tell about a project's security policy, how vulnerabilities are handled / should be reported, and so on. None of those things are mentioned in the file in this pull request. Are you sure you commited the right file?

PR #880, opened 2023-01-06
Fix min-max extrema kernel code

e-kud:
Due to ceiling a size of block, the index may be out of bounds. As index is always used, we must have it valid regardless `while (index < end)` condition.


e-kud:
@pdimov @kylelutz could you please take a look at this patch?

e-kud:
Tentative ping @pdimov

PR #888, opened 2024-08-19
Update sha1.hpp for breaking interface change in Boost.UUID

sehe:
see https://github.com/boostorg/uuid/commit/0f843137a1a479797004f195ec615fdc6ac1c219

The interface is in detail namespace, so the breakage is not on the Boost.UUID maintainers


kylelutz:
Thanks for the patch! Merged similar fix in https://github.com/boostorg/compute/pull/887. Please re-open if this does not resolve your issue.

